PR #6, opened 2015-10-07
Revert "chore(docs): update docs with more placeholder content"

brady-vitrano:
Reverts rackerlabs/zoolander#5


brady-vitrano:
I messed this up because I don't understand this new revert feature :confused:

PR #18, opened 2015-11-13
feat(navigation): first pass at navigation

brady-vitrano:
You can see working example here: 
http://zoolander.rax.io/derek/examples/landing-page.html

This is an initial first pass. As such, it has opportunities for clean up and refactoring.


bighappyface:
+1

Good work

PR #25, opened 2016-03-08
feat(build): Removed Grunt and added build script

SyntaxStacks:


SyntaxStacks:
Large number of changes due to distribution files

scripts/build-dist.js
```diff
@@ -0,0 +1,77 @@
+var _ = require('lodash');
+var harp = require('harp');
```
bighappyface: This isn't used...

Why not `harp compile`?

http://harpjs.com/docs/environment/compile

package.json
```diff
@@ -33,10 +32,13 @@
     "grunt-jscs": "^2.1.0",
     "grunt-newer": "*",
     "grunt-webpack": "^1.0.11",
+    "jquery": "^2.1.4",
     "load-grunt-config": "*",
+    "lodash": "^4.6.1",
+    "node-sass": "^3.4.2",
```
bighappyface: Harp compiles Sass...

SyntaxStacks: `harp.compile` compiles harp projects. If the project contains sass assets, Harp will compile them. I was unable to find a way to compile a single file. Per the documentation harp compile expects paths and not files. Here's the code: https://github.com/sintaxi/harp/blob/master/lib/index.js#L159

We could integrate the scss files into the project itself but the themes currently reside outside of the Harp app and need to be compiled separately from the app. The purpose here was to recreate what Brady was achieving with `grunt build`

SyntaxStacks: If I've missed something in the Harp docs please point me in the right direction. I would much rather rely on Harp then bring in another dependency.

PR #26, opened 2016-03-23
Fix zoolander distribution and workflow

SyntaxStacks:
feat(jade) Replaced EJS with Jade
allows for sane layouts


amberRrucker:
+1!!!!!!!!

PR #32, opened 2016-04-13
feat(zoolander):styling

amberRrucker:
Here we go gang. Some things will adjust with the addition of changes for o365 but mostly there it is. Please let me know if you have feedback. I'll be on slack.


aedile:
@amberRrucker: CTA pattern is missing post CTA button messaging as seen in JIRA task here:
https://jira.rax.io/browse/RSWEB-5411

Basically, in the comp, there's a little slot for a post-message that looks to be a phone number.  Not sure if comps have been updated or we've removed that part.

aedile:
@amberRrucker: Quote pattern feedback based on comparison of the visual comp in ticket - 

1) Icon should be teal. 
2) Quote attribution text should not be all caps. 
3) Quote attribution text should be in italics. 
4) There should be an mdash before the quote attribution. 
5) Pattern implementation has a title in <h2> tags but that's not in the comp.  
6) Example company logo is missing.  

Not sure if these changes are intentional or not.  Just LMK here or in the JIRA task and I'll update accordingly.

styleguide/derek/examples/_solutions.html
```diff
@@ -1,707 +1,600 @@
-<div class="container">
+
+
+
+<!-- main content wrap -->
+<div id="content-wrap" class="styleguide">
```
bighappyface: We can leave out the id and class here. The panels content type template file will contain this tag so we can avoid bringing these over from the legacy markup.

styleguide/derek/examples/_solutions.html
```diff
@@ -1,707 +1,600 @@
...
     </div>
-  </div>
 
-</div>
 
   <!-- Key Offers -->
-<div class="container-fluid river bg-steel-blue">
-    <div class="container">
+  <div class="container-fluid bg-steel-blue">
+    <div class="container key-offers-container standard-padding">
```
bighappyface: We could also extend `.standard-padding` in `.key-offers-container` in the Sass files to save on in-markup class designation, but either way the class setup is semantic and flexible :+1:

PR #33, opened 2016-04-13
fix(mversion): stage precommit build files

bighappyface:


aedile:
Whoops... :grin: 
:+1:

PR #34, opened 2016-04-13
RSWEB-5187 - Move fonticon hosting to cdn.

aedile:
style(font): RSWEB-5187 - Add in fonticon.

fix(fonticon): RSWEB-5187 - Fix container location.


xinbin:
+1

PR #45, opened 2016-04-20
RSWEB-6050: additional resources

jhthompson:
https://jira.rax.io/browse/RSWEB-6050

### DIFF
--- a/styleguide/derek/examples/_office365.html
+++ b/styleguide/derek/examples/_office365.html
@@ -2152,6 +2152,7 @@
     </div>
   </div>
 
+<<<<<<< HEAD
   <!-- Enterprise Managed Services -->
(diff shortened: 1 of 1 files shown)


styleguide/derek/examples/_office365.html
```diff
@@ -2151,4 +2151,59 @@ <h4 class="center">The best part about all of our plans?<br/>
       </div>
     </div>
   </div>
-</div>
\ No newline at end of file
+
+  <!-- Additional Resources -->
+
+  <div class="container-fluid">
+    <div class="container additional-resources-container">
```
amberRrucker: Do we need this class? "additional-resources-container"? Does it have any effect if removed?

SyntaxStacks:
👍

PR #47, opened 2016-04-20
style(cta-markup): polishing up new button styles

ejchiefs:
Replaced button tags with anchor tags and reduced the size of the cta when in sub nav.

### DIFF
--- a/styleguide/_themes/global/scss/_buttons.scss
+++ b/styleguide/_themes/global/scss/_buttons.scss
@@ -1,5 +1,5 @@
 /* the button class can be used for both anchor and input tags, defaults to medium/gray */
-.button, a.button, button{
+.button, button{
   @extend .semi-bold;
   @extend .uppercase;
   @extend .center;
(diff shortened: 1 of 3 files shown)


styleguide/_themes/global/scss/_modules.scss
```diff
@@ -1,28 +1,32 @@
 .subnav{
   background-color:$blue-gray;
   border-radius:0px;
-  button{
-    margin: 10px 0px;
+  .subnav-inside{
+    width: 100%;
+  }
+  button, .button{
```
bighappyface: Why would we ever have a `button` in the subnav?

ejchiefs: Every subnav has a CTA and I remember seeing a conversation about supporting both anchor tags and button tags last week so I was following that decision.

styleguide/_themes/global/scss/_buttons.scss
```diff
@@ -1,5 +1,5 @@
 /* the button class can be used for both anchor and input tags, defaults to medium/gray */
-.button, a.button, button{
+.button, button{
```
bighappyface: I see the CTA examples on `_solutions.html` floating oddly:

screen shot 2016-04-20 at 3 43 25 pm


Need to add `display: block` to this.

bighappyface:
+1 for these adjustments. Will roll another version after the next round of fixes for the CTA.

PR #56, opened 2016-04-21
feat(zoolander):enterprise button up

amberRrucker:
https://jira.rax.io/browse/RSWEB-5478
Wrapped in business tabs. Closing that PR.

### DIFF
--- a/styleguide/_themes/global/scss/_fonts.scss
+++ b/styleguide/_themes/global/scss/_fonts.scss
@@ -55,10 +55,6 @@ a {
     }
 }
 
-ul li, ol li{
-  text-align: left;
-}
-
 blockquote {
   font-size: 1.75rem;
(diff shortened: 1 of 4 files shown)


styleguide/_themes/global/scss/_fonts.scss
```diff
@@ -55,6 +55,9 @@ a {
     }
 }
 
+ul li, ol li{
+  text-align: left;
```
bighappyface: Why do all list items need to be left aligned? Are they being centered by something else?

amberRrucker: Not in zoolander, in www. I'm not sure what is centering it, I haven't had time to delve.

bighappyface:
+1

PR #60, opened 2016-04-22
style(banner-cta): fix banner cta style

ejchiefs:
This PR adds the corrects the CTA styles when applied to banner.


bighappyface:
+1

PR #82, opened 2016-04-28
features pattern creation

amberRrucker:
Creating the features pattern. Let me know if I need to change the padding classes to be more consistent with what we went with for the river and what not. That stuff is more for placement only.


bighappyface:
+1

PR #94, opened 2016-05-02
a fix for the key offers to have equal height

amberRrucker:
Fixed key offers. Sorry guys! Equal height will totally work now.


bighappyface:
+1

PR #95, opened 2016-05-03
refactor(derek): move html to ejs

bighappyface:
This PR removes the unnecessary Jade files in the derek examples and moves the HTML files to EJS for native HTML syntax. This is part of a refactor to isolate out grid pattern partials for easier maintenance.


SyntaxStacks:
👍

pobtastic:
+1

PR #97, opened 2016-05-03
switching up code for resources. adding in and cleaning up

amberRrucker:
switching up code for resources. adding in and hopefully cleaning up …


bighappyface:
+1

PR #102, opened 2016-05-11
Price addition styling needed for o365 tables

amberRrucker:
Styling changes needed for o365 mobile price additions.
https://jira.rax.io/browse/RSWEB-6169

### DIFF
--- a/styleguide/_themes/derek/scss/components/_tabs.scss
+++ b/styleguide/_themes/derek/scss/components/_tabs.scss
@@ -71,7 +71,7 @@
 }
 
 @media only screen and (max-width : 992px) {
-  .tab-content{
+  #myTab .tab-content{
     display:none;
   }
 }
--- a/styleguide/_themes/derek/scss/snowflakes/_o365.scss
+++ b/styleguide/_themes/derek/scss/snowflakes/_o365.scss
@@ -184,4 +184,12 @@
   .check, .times{
     float: right;
   }
+  .pricing .panel-heading{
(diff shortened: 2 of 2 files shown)


bighappyface:
Considering we have been in control of all o365 styling I am concerned we have to introduce several `!important` designations here.

styleguide/_themes/derek/scss/components/_tabs.scss
```diff
@@ -71,7 +71,7 @@
 }
 
 @media only screen and (max-width : 992px) {
-  .tab-content{
+  #myTab .tab-content{
```
SyntaxStacks: `#myTab` could use a better name

and would be better suited as a class.

amberRrucker: Removed those importants. sorry I was debugging a weird preview thing. @SyntaxStacks and I talked about the selector. He's +1. Up to you @bighappyface on merge.

SyntaxStacks:
👍

styleguide/_themes/derek/scss/components/_tabs.scss
bighappyface: In addition to naming I am more concerned with selector specificity. Was this styling setting not working without the id prefixed into the selector?

Sorry if we are splitting hairs on this. I think small details like this are teachable moments for us to determine really good solutions to this stuff instead of the normal CSS/selector onslaught and expansion.

PR #106, opened 2016-05-17
fix(tabs): Fixed style for tabs

SyntaxStacks:
Removed the h2, p stylings to any element nested in the tab anchor would be styled


amberRrucker:
+1

PR #107, opened 2016-05-17
fix(tabs): Tab style updates

SyntaxStacks:
removed h2, p styles in favor of first element


amberRrucker:
+1. You dont really have to do it for the enterprise class but it wont hurt anything either. Go to town Charlie Brown.

PR #109, opened 2016-05-24
RSWEB-6228: Bootstrapped footer

runrog:
https://jira.rax.io/browse/RSWEB-6228

### DIFF
--- a/styleguide/_themes/derek/scss/components/_footer.scss
+++ b/styleguide/_themes/derek/scss/components/_footer.scss
@@ -1,1 +1,102 @@
+.footer-menu-link {
+  font-size: 1.2rem;
+  color: darken($light-gray, 15%);
+}
 
+.footer-tab-link {
+  font-size: 1.3rem;
+  text-transform: uppercase;
+  color: $white;
+  &:hover, &:focus {
+    cursor: default;
+    text-decoration: none;
+    color: $white;
+  }
+}
+
+.footer-menu {
+  list-style: none;
+  padding: 0;
+  margin: 0;
+}
+
+.footer-tabs {
+  list-style: none;
+  padding: 0;
+  margin-bottom: 25px;
+}
+
+.footer-menu-split {
+  column-count: 3;
+}
+
+.footer-menu-item {
+  list-style: none;
+  line-height: 25px;
+}
+
+@media only screen and (max-width: $screen-sm) {
+  .footer-menu-link {
+    display: block;
+    background-color: lighten($blue-gray, 5%);
+    padding: 15px;
+    margin: 0;
+    &:hover {
+      background-color: lighten($blue-gray, 8%);
+    }
+  }
+  .footer-tabs {
+    margin-bottom: 0;
+  }
+  .footer-tab-link {
+    padding: 15px;
+    display: block;
+    &:hover {
+      cursor: pointer;
+    }
+  }
+  .footer-menu {
+    text-align: center;
+  }
+  .footer-menu-split {
+    column-count: 1;
+  }
+  .footer-menu-item {
+    margin: 0;
+  }
+
+  // Nesting panel selectors to fix bootstrap-related maintainable css issue.
+  .footer {
+    padding: 0;
+    .panel-body {
+      padding: 0;
+      border-top: none !important;
+    }
+    .panel-group {
+      .panel + .panel {
+        margin-top: 0;
+      }
+    }
+    .panel-default {
+      border: 0;
+    }
+    .panel-heading {
+      border-top-right-radius: 0;
+      border-top-left-radius: 0;
+      padding: 0;
+    }
+    .panel-title {
+      background-color: $blue-gray;
+      border-bottom: 1px solid lighten($blue-gray, 5%);
+      text-align: center;
+      padding-left: 0;
+      padding-right: 0;
+    }
+    .panel-title a {
+      color: $white;
+    }
+    .panel-body a {
+      color: darken($light-gray, 15%);
+    }
+  }
+}
--- a/styleguide/derek/_partials/_footer.ejs
+++ b/styleguide/derek/_partials/_footer.ejs
@@ -1,5 +1,56 @@
 <!--[footernav]-->
-<nav class="">
+<div class="bg-blue-gray standard-padding-full footer">
+    <div class="container">
+        <div class="row">
+            <ul id="footer-accordion" class="footer-tabs">
+                <li class="col-sm-6"><a href="#footer-tab-1" class="footer-tab-link" data-toggle="tab">About Rackspace</a></li>
+                <li class="col-sm-2"><a href="#footer-tab-2" class="footer-tab-link" data-toggle="tab">Blog</a></li>
+                <li class="col-sm-2"><a href="#footer-tab-3" class="footer-tab-link" data-toggle="tab">Support</a></li>
+                <li class="col-sm-2"><a href="#footer-tab-4" class="footer-tab-link" data-toggle="tab">Corporate</a></li>
+            </ul>
 
-</nav>
+            <div id="footer-tab-1" class="col-sm-6">
+                <ul class="footer-menu footer-menu-split">
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Leadership</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Legal</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Contact Information</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Events</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Customer Stories</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Marketplace</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Responsibility</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">News</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Investors</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Careers</a></li>                    
+                </ul>
+            </div>
+
+            <div id="footer-tab-2" class="col-sm-2">
+                <ul class="footer-menu">
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">The Rackspace Blog</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Developer Blog</a></li>
+                </ul>
+            </div>
+
+            <div id="footer-tab-3" class="col-sm-2">
+                <ul class="footer-menu">
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Knowledge Center</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Support Forums</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">API Documentation</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Support Home</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Developer Center</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Partner Locator</a></li>
+                </ul>
+            </div>
+
+            <div id="footer-tab-4" class="col-sm-2">
+                <ul class="footer-menu">
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Privacy Statement</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Website Terms</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Trademarks</a></li>
+                    <li class="footer-menu-item"><a class="footer-menu-link" href="#">Sitemap</a></li>
+                </ul>
+            </div>
+        </div>
+    </div>
+</div>
 <!--[/footernav]-->
--- a/styleguide/js/zoolander.js
+++ b/styleguide/js/zoolander.js
@@ -21,7 +21,7 @@ jQuery(document).ready(function ($) {
     jQuery ('.title a').click(function (event) {
       jQuery (this).parent().parent().find('.fa-arrow-right').animate({  borderSpacing: +90 }, {
       step: function(now,fx) {
-          $(this).css('-webkit-transform','rotate('+now+'deg)'); 
+          $(this).css('-webkit-transform','rotate('+now+'deg)');
           $(this).css('-moz-transform','rotate('+now+'deg)');
           $(this).css('transform','rotate('+now+'deg)');
           counter = 1;
@@ -29,6 +29,7 @@ jQuery(document).ready(function ($) {
       duration:'slow'
       },'linear');
     });
+
+    $('#footer-accordion').tabCollapse();
+
 });
-
-
--- a/styleguide/_themes/global/scss/_modifiers.scss
+++ b/styleguide/_themes/global/scss/_modifiers.scss
@@ -32,4 +32,3 @@ img { max-width: 100%; }
 .black {color:$black;}
 .darkest-gray{color:$darkest-gray;}
 .blue-gray{color:$blue-gray;}
-


styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,117 @@
 .footer{
-  
-}
\ No newline at end of file
+  color:#fff;
```
amberRrucker: Use swatches for colors. example $white. Look at the swatches scss file for those names.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,117 @@
...
+    }
+  }
+}
+/*  Mobile   */
+@media only screen and (min-width:50px) and (max-width: 767px)  {
+  .fluid-footer-container{
+    padding: 0px 0px;
+  }
+  .footer{
+    width: 100%;
```
amberRrucker: extend .full-width;

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,117 @@
...
+    padding: 0px 0px;
```
amberRrucker: For all these use extend .no-padding; or extend .no-margin;

WidgetsBurritos:
Just ran it locally.  Can you make the accordion collapse other sections when expanding another section?

WidgetsBurritos:
Also I was initially wrong about needing `row` and `clearfix` classes.  I had some other issue that necessitated that, but it works fine without those now.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,117 @@
...
\ No newline at end of file
+  color:#fff;
+  ul{
+    list-style: none;
+    padding: 0px;
+    margin:0px;
+    li{
+      padding: 0px;
+      margin:5px 0px;
+      a{
```
WidgetsBurritos: Can we a `list-style: none;` inside the li here as well?

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,117 @@
...
+            cursor: hand;
+          }
+        }
+        ul{
+          li{
+            ul{
+              li{
+                padding:0px;
+                border-bottom: 1px solid #3b4047;
+                a{
```
bighappyface: This tree of selectors is out of hand and demonstrates why classes are the superior selector, not just in performance, but in authoring.

`.footer ul li ul li ul li a` is the final selector. In addition to being less-than-desireable CSS, it reflects what is probably hairy markup.

`.footer-menu-link` or something to that effect would be much cleaner, performant, and maintainable.

Are we using something already available in Bootstrap, or are we making it all from scratch?

WidgetsBurritos: I'd tend to agree with this.  

@runrog I can force drupal to output depth classes on the <li>'s.  This would probably be sufficient in zoolander as well.  

Instead of that `ul li ul li ul li` nesting, you can just do `ul li.depth-3`.  That should help cleanup that css a bit.

styleguide/js/zoolander.js
```diff
@@ -21,14 +21,41 @@ jQuery(document).ready(function ($) {
...
 
+    mobileFooterCheck();
+    $(window).resize(function(){
+      mobileFooterCheck();
+    });
+    function mobileFooterCheck(){
+      var footerSectionId = 0;
+      $('.footer ul li span.menu-title').each(function(){
+        footerSectionId++;
+        var _this = $(this);
```
WidgetsBurritos: It's a common practice to name a variable referencing a jQuery wrapped object with a $, because it helps the person reading the code know you are dealing with jQuery and not a common JavaScript variable. 

I'd recommend `var $this = $(this);` instead of `var _this = $(this);`

See: http://stackoverflow.com/questions/6209462/when-why-to-prefix-variables-with-when-using-jquery

bighappyface:
It looks like this footer is all custom, CSS to JS. Do we have options within OOTB Bootstrap that we can explore? If we could minimize effort/code on the footer we might save some headaches.

runrog:
@bighappyface The markup is utilizing bootstrap rules as much as possible (cols/rows etc.) the only real sass added was to stylize the ul's. Bootstrap doesn't have anything built in that does what we need to for ul's. Also, agreed on the ul nesting rules. Will added classes so it's easier to read/maintain. 

The JS is basically dynamically adding the bootstrap attributes for the collapsable accordion on mobile. So that it only works on mobile and then get's removed if on desktop. 

I am revisiting these issues today and will resubmit. thanks.

styleguide/js/zoolander.js
```diff
@@ -21,14 +21,41 @@ jQuery(document).ready(function ($) {
...
+        var attr = _this.attr('data-toggle');
+        //tells us if screen is tablet & or mobile
+        if ($('.footer ul li').css('text-align')=='center') {
```
WidgetsBurritos: I'm not sure checking text alignment is the best way to identify viewport. If for whatever reason it ever becomes left-aligned we'd have to go back and fix this functionality. 

Perhaps something like this instead? (Note: This is not a definitive solution, but more of something open for discussion). 
http://stackoverflow.com/questions/18575582/how-to-detect-responsive-breakpoints-of-twitter-bootstrap-3-using-javascript

bighappyface:
@runrog I mean something similar to the collapsable tab stuff that has a more formal opinion on how to do navs.

Despite being in bootstrap it still looks like this is totally custom.

styleguide/derek/_partials/_footer.ejs
```diff
@@ -1,5 +1,65 @@
 <!--[footernav]-->
-<nav class="">
-
+<nav class="container footer">
+    <ul class="row clearfix">
+        <li class="col-xs-12 col-sm-6"><span class="menu-title">About Rackspace</span>
```
WidgetsBurritos: So I think this may have been my fault, but I don't think this col-xs-12's are necessary.  These <li>'s span the full width by default anyway, and it will only go to 6 columns on `sm` and up.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,72 @@
 .footer{
-  
-}
\ No newline at end of file
+  .col-sm-12{
+      padding-left: 0px;
+  }
```
amberRrucker: no altering bootstrap base elements sir.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,72 @@
...
+  .footer{
+    padding:0px 0px;
+    .container{
+      padding:0px 0px;
+    }
+    .col-sm-12{
+      padding: 0px 0px;
+    }
+    h6{
+      background-color:darken( $blue-gray, 05%);
```
amberRrucker: Just use blue gray. no darken or lighten please.

amberRrucker:
Very good. Much better. I made a couple comments @runrog so just quick clean up those and youre golden in my opinion.

runrog:
@amberRrucker no probs. I'll make those changes

WidgetsBurritos:
@runrog Changes look good.  Only issue I still see is when in `sm` mode, you have a scroll bar on the bottom of the page.   

image

I'm wondering if it would make sense to which all the col-sm-* to col-md-* and then change the isMobile() function to isDesktop() and check if you're greater than 768 instead? I think creating exception rules for desktop instead of mobile, fits the mobile-first paradigm anyway.  

I'm not sure if anything else on the site uses that accordion functionality, so we'd probably need to double check on that. 

Any thoughts on that @bighappyface or @amberRrucker?

styleguide/derek/_partials/_footer.jade
```diff
@@ -1,2 +1,2 @@
-.container-fluid.bg-blue-gray.standard-padding
+.container-fluid.bg-blue-gray.fluid-footer-container
```
bighappyface: We can lose this file and add `<div class="container-fluid bg-blue-gray fluid-footer-container">...</div>` to `_footer.ejs`

amberRrucker: This file will have the rug and basement as well.

bighappyface: @amberRrucker won't that get confusing?

I think it would be better to have partials included in the layout like `_footer` is as they are all distinct:

https://github.com/rackerlabs/zoolander/blob/697650c11fa5c433bbf4639e6b70db7e2a697a3d/styleguide/derek/examples/_layout.jade#L6

amberRrucker: @bighappyface I included it like the navigation file includes the ceiling and mainnav partials. Since the footer wont ever appear without the rug and basement I felt like it would be easier to include them within that page as a group but have them in their own partials.

bighappyface: I don't really agree with the way `_navbar` includes the `_ceiling` and `_mainnav` either, but at least `_navbar` and `_mainnav` are distinct names and indicative of what they are.

Having two files named `_footer`, regardless of the different extensions that have virtually no semantic meaning to what they are in the styleguide, is confusing, and confusing is not "easier."

Let's consolidate those files now so that we don't have to do it in the future.

amberRrucker: @bighappyface thats fine with me. My only concern is on landing pages the ceiling is included and not the mainnav. In the future we're talking about including the rug on landing pages but not the rest of the site. Allowing the flexibility for including those on different "templates" and not on others would be nice.

bighappyface: @amberRrucker that's a problem for Drupal, not this styleguide. We should solve Drupal problems in Drupal, and HTML/CSS problems here.

styleguide/js/zoolander.js
```diff
@@ -21,14 +21,52 @@ jQuery(document).ready(function ($) {
...
+      return typeof window.matchMedia !== 'undefined' ? window.matchMedia("only screen and (max-width: 768px)").matches : false;
+    }
+    //Additional viewport check (replacing isMobile)
+    function isDesktop() {
+      return typeof window.matchMedia !== 'undefined' ? window.matchMedia("only screen and (min-width: 992px)").matches : false;
+    }
+    /**
+     * Simple accordion logic for use on mobile displays
+     */
+    Rackspace.Accordion = (function($) {
```
bighappyface: This is the first instance of custom JS that is in Zoolander that will be brought into www. I would prefer not to copy/paste it into www and I don't think the RJL is a good place for it, so we should look at putting this into it's own file in using a `script` element to bring it in prior to this `ready` block. Once it is in it's own file we can integrate it into www with gulp and add the `Rackspace.Accordion.init();` line in `rackspace.js`.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,74 @@
...
-  
-}
\ No newline at end of file
+  .accordionTitle h6{
+    color: $white;
+    font-size: 1.3rem;
+  }
+  ul{
+    list-style: none;
+    @extend .no-padding;
```
bighappyface: I propose we use `padding: 0` and `margin: 0` here in favor of the class extension. In addition to being `sass-lint` compliant (placeholders are preferred for extension), we will save a few bytes in text, but more importantly, gain a more flexible design by not being coupled to those classes for simple CSS properties.

`@extend` is fine but we should be more conservative/tactical with it, and if we are to use those classes, we should apply the classes on the elements directly instead of extending it here. Also, because we can't extend classes within media queries, the use of `@extend` is limited, as is seen below in numerous `padding` resets. It will be more maintainable and consistent to just set the properties.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,74 @@
...
+    @extend .no-margin;
+    li{
+      @extend .no-padding;
+      margin:5px 0px;
```
bighappyface: Please drop units for 0 values - e.g. `margin:5px 0;`

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,74 @@
...
+  }
+  .footer{
+    padding:0px 0px;
+    .accordionTitle{
+      background-color: $blue-gray;
+      border-bottom:1px solid lighten($blue-gray, 05%);
+      h6{
+        text-align: center;
+        padding:12px;
+        margin: 0px 0px;
```
bighappyface: Please drop units for 0 values - e.g. `margin: 0`

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,74 @@
...
+}
+/*Mobile*/
+@media only screen and (min-width:768px) and (max-width: 991px){
+  .full-width-small{
+    width:100%;
+  }
+}
+@media only screen and (min-width:50px) and (max-width: 992px){
+  .fluid-footer-container{
+    padding:0px 0px;
```
bighappyface: Please drop units for 0 values - e.g. `padding: 0`

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,74 @@
...
+  }
+  .footer{
+    padding:0px 0px;
```
bighappyface: Please drop units for 0 values - e.g. `padding: 0`

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,74 @@
...
+    }
+    .accordionMenu{
+      background-color: lighten($blue-gray, 05%);
+    }
+    ul{
+      &.split{
+        column-count: 1;
+      }
+      li{
+        margin:0px 0px;
```
bighappyface: Please drop units for 0 values

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,74 @@
 .footer{
-  
-}
\ No newline at end of file
+  .accordionTitle h6{
```
bighappyface: It would be more modular to rename this class to `.footer-accordion-title` and drop the `h6`. That way we can apply that class directly to whatever the heading/text element is without breaking anything if we need/decide to change it. That class name will also pull it out of being nested in `.footer` making the selector more efficient.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,74 @@
...
+        padding:12px;
+        margin: 0px 0px;
+      }
+      &:hover{
+        background-color: darken( $blue-gray, 05%);
+        cursor:pointer;
+        cursor: hand;
+      }
+    }
+    .accordionMenu{
```
bighappyface: Similar to the comment above on `.accordionTitle` this could be a standalone class `.footer-accordion-menu` that is more efficient and modular.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,74 @@
...
+    .accordionTitle{
+      background-color: $blue-gray;
+      border-bottom:1px solid lighten($blue-gray, 05%);
+      h6{
+        text-align: center;
+        padding:12px;
+        margin: 0px 0px;
+      }
+      &:hover{
+        background-color: darken( $blue-gray, 05%);
```
bighappyface: Are we using `darken`or do we favor fixed color values? I'm fine with mixins like this but prefer consistency over convenience.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,74 @@
...
+      }
+    }
+    &.split{
+      column-count: 3;
+    }
+  }
+}
+/*Mobile*/
+@media only screen and (min-width:768px) and (max-width: 991px){
+  .full-width-small{
```
bighappyface: This is out of place.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,87 @@
...
+    list-style: none;
+    padding: 0;
+    margin: 0;
+  }
+  .footer-menu {
+    li {
+      list-style: none;
+      padding: 0;
+      margin: 8px 0;
+      a {
```
WidgetsBurritos: We probably should either move this a outside of the li (directly under .footer-menu), or perhaps even assign classes to the a links and take it out of .footer-menu altogether. Let's get @bighappyface's feedback on that.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,87 @@
...
+    }
+  }
+  .footer-accordion-title {
+    color: $white;
+    font-size: 1.3rem;
+  }
+  .footer-tabs {
+    li{
+      a{
+        color:$white;
```
WidgetsBurritos: same as above here.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,87 @@
...
+      }
+    }
+  }
+}
+@media only screen and (max-width: $screen-sm) {
+  .fluid-footer-container{
+    padding: 0 0;
```
WidgetsBurritos: I'd say `padding: 0;` would preferable to `padding: 0 0;`

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,87 @@
 .footer{
-  
-}
\ No newline at end of file
+  ul{
+    list-style: none;
+    padding: 0;
+    margin: 0;
+  }
+  .footer-menu {
```
WidgetsBurritos: You should be able to pull `.footer-menu{}` to the top level instead of nested inside `.footer`

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,87 @@
...
+    }
+  }
+}
+@media only screen and (max-width: $screen-sm) {
+  .fluid-footer-container{
+    padding: 0 0;
+  }
+  .footer{
+    padding:0 0;
+    .footer-menu {
```
WidgetsBurritos: Pull the `.footer-menu` out of the `.footer` inside the media query as well.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,87 @@
...
+        display: block;
+        background-color: lighten($blue-gray, 05%);
+        padding:15px;
+        margin: 0 0;
+        &:hover{
+          background-color: lighten($blue-gray,08%);
+        }
+      }
+      li{
+        margin: 0 0;
```
WidgetsBurritos: Same as `padding: 0;` issue.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,87 @@
...
+      }
+    }
+    .panel-body{
```
WidgetsBurritos: Since all the .panel-* are mandated by tab collapse, this should remain inside .footer, but I think anything else, should be in the top-level.

styleguide/derek/_partials/_footer.ejs
```diff
@@ -1,5 +1,53 @@
 <!--[footernav]-->
-<nav class="">
+<div class="container-fluid bg-blue-gray standard-padding footer">
+    <div class="container">
+        <div class="row">
+            <ul id="footer-accordion" class="footer-tabs">
+                <li class="col-sm-6 active"><a href="#footer-tab-1" data-toggle="tab">About Rackspace</a></li>
```
WidgetsBurritos: Maybe we should take `active` off of here.  The current script starts out collapsed, so maybe it makes sense to do the same.

styleguide/_themes/derek/scss/components/_footer.scss
bighappyface: I am always in favor of styling classes over elements. If we are good with Drupal then I say take the plunge.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,77 @@
...
+    padding:15px;
+    margin: 0;
+    &:hover{
+      background-color: lighten($blue-gray,08%);
+    }
+  }
+  a.footer-tab-link{
+    padding:15px;
+    display: block;
+    color:$white;
```
WidgetsBurritos: You shouldn't have to redefine color here if it's already set as the default color for the same selector outside of the media query.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,77 @@
...
+    margin: 0;
+  }
+}
+.footer-menu {
+  li {
+    list-style: none;
+    padding: 0;
+    margin: 8px 0;
+  }
+  &.split {
```
WidgetsBurritos: Maybe create a `.footer-menu-split` class instead and take out of here.  `.split` is pretty generic, so it'd be ideal to have something a bit more specific to the context.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,77 @@
...
-}
\ No newline at end of file
+  ul{
+    list-style: none;
+    padding: 0;
+    margin: 0;
+  }
+}
+.footer-menu {
+  li {
```
WidgetsBurritos: Should these li's to `.footer-menu-item`

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,77 @@
...
+}
+a.footer-tab-link{
+  color:$white;
+  font-size: 1.3rem;
+}
 .footer{
-  
-}
\ No newline at end of file
+  ul{
```
WidgetsBurritos: I'd rename this `.footer-tabs` and then just remove list style, padding, and margin from `.footer-menu` separately.

WidgetsBurritos:
:+1:

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1 +1,87 @@
...
+    column-count: 3;
+}
+li.footer-menu-item {
+  list-style: none;
+  line-height: 25px;
+}
+@media only screen and (max-width: $screen-sm) {
+  a.footer-menu-link{
+    display: block;
+    background-color: lighten($blue-gray, 05%);
```
amberRrucker: just blue-gray man. no lighten for the bg color por favor. it wont match the other elements.

amberRrucker:
+1

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1 +1,87 @@
-
+a.footer-menu-link {
```
bighappyface: @runrog this is almost there. Please remove the element prefixes here and elsewhere in this file. A class definition does not need an element prefix.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1 +1,87 @@
...
+  color: darken($light-gray, 15%);
+  font-size: 1.2rem;
+}
+a.footer-tab-link{
+  color:$white;
+  font-size: 1.3rem;
+  text-transform: uppercase;
+  &:hover{
```
WidgetsBurritos: Can you change this to `&:hover, &:focus {` to this too?

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1 +1,87 @@
...
+      border:0;
+    }
+    .panel-heading{
+      border-top-right-radius: 0;
+      border-top-left-radius: 0;
+      padding:0;
+    }
+    .panel-title{
+      background-color: $blue-gray;
+      border-bottom: 1px solid lighten($blue-gray, 05%);
```
bighappyface: 05%?

runrog: @bighappyface fixed. For some reason, I was thinking in double digits

WidgetsBurritos:
:+1: for the latest revisions.

bighappyface:
+1

bighappyface:
Welcome to Zoolander @runrog!

https://github.com/rackerlabs/zoolander/graphs/contributors

runrog:
@bighappyface thanks! I feel like a real dev now 😝

WidgetsBurritos:
@runrog great job on all of this! Thanks for putting up with our constant changes and nitpicking.

PR #110, opened 2016-05-25
Global elements - ceiling

amberRrucker:
Creating the bootstrapped ceiling.

### DIFF
--- a/styleguide/_themes/derek/scss/components/_ceiling.scss
+++ b/styleguide/_themes/derek/scss/components/_ceiling.scss
@@ -12,12 +12,12 @@ $ceiling-font-size: 10px;
   @extend .visible-lg;
   @extend .uppercase;
   @extend .semi-bold;
-
+  @extend .no-border;
   margin: 0;
   min-height: $line-height-computed;
-  background: white;
+  background: $white;
   font-size: $ceiling-font-size;
-  border: none;
+  border-radius: 0px;
 
   a {
     font-size: $ceiling-font-size;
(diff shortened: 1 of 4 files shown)


styleguide/_themes/derek/scss/components/_ceiling.scss
```diff
@@ -44,13 +46,11 @@ $ceiling-font-size: 11px;
...
 
     > li{
-      margin: $ceiling-margin 0;
-      padding: 0;
-      &:after {
-        content: "|";
-      }
-      &:last-child::after {
-        content: " ";
+      margin: floor($ceiling-margin + 2) 0;
```
WidgetsBurritos: Is this floor necessary? Isn't ceiling-margin already an integer, based on its definition already using floor(): 
```$ceiling-margin: floor($navbar-margin-bottom / 4);```

I suppose it doesn't hurt anything as it all gets compiled down anyway, but just thought I'd point it out.

Comment (amberRrucker) [inline styleguide/_themes/derek/scss/components/_ceiling.scss]: @WidgetsBurritos it's basically a just in case an integer gets changed or we're dividing by something that results in a non whole number. A safety net. @bighappyface what do you think? leave it in as a just in case or is it unnecessary?

Comment (WidgetsBurritos) [inline styleguide/_themes/derek/scss/components/_ceiling.scss]: Fair enough. It's probably a non-issue honestly since it doesn't affect the final CSS one way or another.

Comment (bighappyface) [inline styleguide/_themes/derek/scss/components/_ceiling.scss]: I can live with the redundant `floor`

**Comment by WidgetsBurritos** [comment]
+1

File: styleguide/derek/_partials/_rug.ejs [hunk 2]
```diff
@@ -0,0 +1,12 @@
+<!--[rug]-->
```
WidgetsBurritos: Did you intentionally include the rug in this PR?

PR #111, opened 2016-05-31
Rug Partial and Global Element Reorg

amberRrucker:
https://jira.rax.io/browse/RSWEB-6227


styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,15 @@
-.footer{
-  
-}
\ No newline at end of file
+.rug{
```
bighappyface: This should be in `rug.scss` as there is a lot of work going into this file in #109

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,15 @@
...
+  .text37{
```
bighappyface: I favor a more modular class name here like `.rug-header` or `.rug-text-large` that isn't nested.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,15 @@
...
+    font-size: 37px;
+    line-height: 24px;
+  }
+  &, *{
+    @extend .white;
+  }
+  .contactLink{
```
bighappyface: A more module name would be better here too. A single class is enough to define this property instead of nested class names.

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,3 +1,15 @@
-.footer{
-  
-}
\ No newline at end of file
+.rug{
+  .text37{
+    font-size: 37px;
+    line-height: 24px;
+  }
+  &, *{
```
bighappyface: There has got to be a better way to make the text white than this

bighappyface:
The Jade/EJS migrations here are a step in a good direction. Fewer files that contain what is expected :)

styleguide/_themes/derek/scss/components/_footer.scss
amberRrucker: I can add class="white" to each of these elements but since everything in the rug will always be white, this seems a lot less hefty.

styleguide/_themes/derek/scss/components/_footer.scss
amberRrucker: you got it.

styleguide/_themes/derek/scss/components/_footer.scss
amberRrucker: @bighappyface You want the rug, footer menu and basement in their own files? Those two things have very few lines styling wise, I think it's a little less confusing to have those all in one file.

styleguide/_themes/derek/scss/components/_footer.scss
bighappyface: If we put all items that we want near the bottom of the page in this file then I am unclear of the where to draw the line on what goes in it's own file and what does not. If we adhere to a modular CSS design then each "module" goes in it's own file as a simple rule of thumb as files are cheap and organization is more important than saving files.

I see the rug as a module and it's position on the page isn't relevant to the HTML/CSS/styleguide file structure.

amberRrucker: @bighappyface I made the changes to the PR we talked about except this one. I'm not sure what is a better way to make all the text within the rug white aside from adding the class white to every element. This way seems to be less repetitive and more encompassing. Do you have any alternatives?

styleguide/derek/examples/homepage.ejs
```diff
@@ -303,11 +303,11 @@
                 </div>
               </div>
             </div>
-            <div class="col-md-6 col-md-push-2">
```
bighappyface: I removed this `.col-md-push-2` to fix the extra 2-column push beyond the 12 units that were already covered by the two `.col-md-6` divs.

styleguide/derek/_partials/_rug.ejs
```diff
@@ -1,12 +1,47 @@
 <!--[rug]-->
-<div class="container rug">
```
bighappyface: @amberRrucker you will see here and below that I refactored the rug to use a few modular classes to accomplish the desired results. Please feel free to adjust it further for aesthetics.

The key lesson here is the use of single-depth class selectors that do not conflict with any other styles, do not extend any other styles, and express their usage in a semantic fashion. Such class design fits in very well with the classes provided by Bootstrap.

styleguide/derek/_partials/_rug.ejs
```diff
@@ -1,12 +1,47 @@
...
+<div class="container-fluid bg-blue-gray half-padding">
+<div class="container hidden-xs hidden-sm">
   <div class="row">
     <div class="col-sm-3">
-      <div class="col-sm-3">
-        <h5>We've got your back</h5>
-        <h4 class="text34">24x7x365</h4>
+      <h5 class="rug-subhead-medium no-margin">We've got your back</h5>
+      <h4 class="rug-subhead-large no-margin">24<small class="white">x</small>7<small class="white">x</small>365</h4>
```
bighappyface: Not a fan of using `.white` here to override colors set for all `small` elements within `h4` but that is what a flexible modifier class provides. In the future we need to refactor our CSS to not style elements but instead style classes and let styles cascade properly. In this case, if the `small` elements didn't have styles applied to them then they could inherit their color from `.rug-subhead-large`. Just because something is `small` doesn't mean it has to have a specific color, so it is better to leave styling to the outer element or apply a class for _explicit_ styling when needed.

styleguide/derek/_partials/_rug.ejs
```diff
@@ -1,12 +1,47 @@
...
     </div>
+    <div class="col-sm-3">
+      <div class="row">
+        <div class="col-lg-2 col-md-3">
+          <i class="fa fa-mobile fa-4x rug-icon"></i>
+        </div>
+        <div class="col-lg-10 col-md-9 rug-contact-link">
+          <a class="sales-phone rug-contact-link" href="tel:1-844-858-5963">1-844-858-5963</a> Sales<br/>
```
bighappyface: Same thing here... just because text needs to be made an `a` doesn't mean it has to share the same color as all other anchors, nor does it mean we have to add in explicit overrides via class assignment or CSS. It's best to leave format/functionality elements alone and style by class as needed, but for now this will work.

styleguide/derek/_partials/_rug.ejs
```diff
@@ -1,12 +1,47 @@
...
-      <div class="col-sm-3">
-        <h5>We've got your back</h5>
-        <h4 class="text34">24x7x365</h4>
+      <h5 class="rug-subhead-medium no-margin">We've got your back</h5>
+      <h4 class="rug-subhead-large no-margin">24<small class="white">x</small>7<small class="white">x</small>365</h4>
     </div>
+    <div class="col-sm-3">
+      <div class="row">
+        <div class="col-lg-2 col-md-3">
+          <i class="fa fa-mobile fa-4x rug-icon"></i>
```
bighappyface: Applying a class making it clear this is an icon in the rug helps identify it for both CSS and JS (if needed). The `.rug-icon` class has a `color: $white` and applying the class changes the color. No super selector (`.rug *`) is needed to color this element.

styleguide/_themes/derek/scss/components/_basement.scss
```diff
@@ -0,0 +1,5 @@
+.basement{
+  &, *{
```
bighappyface: Will the basement be continued in another PR? If not this should be refactored now.

styleguide/derek/_partials/_rug.ejs
amberRrucker: Totally included that by accident. It does not need to be there. Sorry!

bighappyface: @amberRrucker I added it there to make the text white. Not a fan of the need but it works and requires no new CSS.

amberRrucker: @bighappyface Is there colors set on small elements within h4s? That seems very weirdly specific. The rule I used should just make everything within the rug white? Can we chat about this today at some point? I may not be understanding why were setting the color this way, it seems clunky.

bighappyface: Yes, the color for small in headers is being set:

```css
// derek.css
h1 small,h1 .small,h2 small,h2 .small,h3 small,h3 .small,h4 small,.rug-to-top a small,#top-link-desktop a small,h4 .small,.rug-to-top a .small,#top-link-desktop a .small,h5 small,h5 .small,h6 small,h6 .small,.h1 small,.h1 .small,.h2 small,.h2 .small,.h3 small,.h3 .small,.h4 small,.h4 .small,.h5 small,.h5 .small,.h6 small,.h6 .small {
    font-weight: normal;
    line-height: 1;
    color: #888c92
}
```

WidgetsBurritos: So not to throw a wrench into the equation, but Drupal counterpart of this is going to involve the `t()` function. Generally speaking, its usually best to avoid passing too much HTML into `t()`, unless absolutely necessary. Can we just remove the text-transform on `.rug-subhead-large` instead?

pobtastic: We won't need to translate **24x7x365** I've a feeling it's pretty universal!

amberRrucker: @WidgetsBurritos waiting on a brand response for this one to see if we need to find an alternative solution.

WidgetsBurritos: @pobtastic So are you suggesting not passing it through t() at all, or just that the translation for it won't matter? 24x7x365 is currently passed into t() so that's why I bring it up.

pobtastic: Yeah I wouldn't bother translating it, numbering like that is pretty much universal ~ plus it's kinda like a slogan, and they aren't usually translated either.

bighappyface:
Closed in favor of #114

PR #112, opened 2016-06-01
Adjusting tabs structure per rsweb-6236

amberRrucker:
Changed bg color and added a container to the interior of the tabs per EJ's request to fix issues on cloud pricing pages.


bighappyface:
+1

PR #115, opened 2016-06-02
RSWEB-6403: ias clickarea update

SamuelRandall:
https://jira.rax.io/browse/RSWEB-6403

Refactoring ias to no longer use hover state if url is not present.

### DIFF
--- a/styleguide/_themes/derek/scss/_modifiers.scss
+++ b/styleguide/_themes/derek/scss/_modifiers.scss
@@ -16,20 +16,30 @@ $body-bg:  #fff !default;
 .bg-crumpled-paper { background-color: $light-gray; background-image: url(https://752f77aa107738c25d93-f083e9a6295a3f0714fa019ffdca65c3.ssl.cf1.rackcdn.com/elements/bg-crumpled-paper.png); }
 
 //centered-divs
-.center-block{ margin: 0 auto; float:none;}
+.center-block { margin: 0 auto; float: none; }
 
 //equal height columns
-.row-eq-height {
+.row-eq-height, .flex-container {
   display: -webkit-box;
   display: -webkit-flex;
   display: -ms-flexbox;
   display:         flex;
 }
 
+.flex-item {
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display:         flex;
+  & > * {
+    width: 100%;
+  }
+}
+
 @media only screen and (max-width : 770px) {
   //equal height columns remove in small
-  .row-eq-height {
-  display: block;
+  .row-eq-height, .flex-container {
+    display: block;
   }
 }
 
--- a/styleguide/derek/_partials/grid/widget/ias-no-hover.ejs
+++ b/styleguide/derek/_partials/grid/widget/ias-no-hover.ejs
@@ -2,7 +2,7 @@
 <% if(!desc) { var desc = "This description cannot be longer than 75 characters including all spaces."; } %>
 <!-- Infrastructure and Services Pattern -->
 
-<a class="ias ias-hover" href="http://www.google.com">
+<div class="ias">
   <div class="row">
     <div class="col-sm-4 col-xs-3 ias-img">
       <img src="http://placehold.it/180x180">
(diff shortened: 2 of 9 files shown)


jhthompson:
Hey @bighappyface I know we spoke about possibly revisiting the equal height beyond this method but I went ahead and updated my work here to show a working version.

When making the changes I noticed that the 2 up/2 down or "4 Column" section was setting the img off center because the img was not large enough to fill the col div wrapping it. This specific layout could probably use separate markup from the others. The img is being stretched beyond the size displayed in the other variations.

To show the pattern working correctly I have added an alternate placeholder img that fills the col. I will be glad to go back and clean that out in the event that we use this equal height method beyond the current PR.

Also, reluctantly I used the * to make sure the contents of the flex-item fills the col. It makes sense to me to use it here but I could be missing issues it might cause. Thanks!

styleguide/derek/_partials/grid/widget/ias-no-hover.ejs
```diff
@@ -0,0 +1,18 @@
+<!-- Default description to allow for partial override to demonstrate really long text and equal height -->
+<% if(!desc) { var desc = "This description cannot be longer than 75 characters including all spaces."; } %>
+<!-- Infrastructure and Services Pattern -->
+
+<div class="ias">
+  <div class="row">
+    <div class="col-sm-4 col-xs-3 ias-img">
```
jhthompson: @bighappyface do you think maybe the class here should be something else since it doesnt directly alter the img tag itself? Maybe something like ```ias-img-col``` or ```ias-img-container``` ?

styleguide/_themes/derek/scss/_modifiers.scss
```diff
@@ -16,20 +16,30 @@ $body-bg:  #fff !default;
...
   display: -ms-flexbox;
   display:         flex;
 }
 
+.flex-item {
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display:         flex;
+  & > * {
```
SamuelRandall: @bighappyface @writes The sass linter won't allow this, I'll need help finding an alternate solution.

WidgetsBurritos: Is the objection with the vendor prefixes or the `& > *` wildcard selector?

In reading up on why the sass linter recommends against using vendor prefixes, it appears to recommend using a CSS post processor (Autoprefixer) or mixins from Compass or Boubon in addition to a pre processor (Sass). 

More info:
https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#vendorprefix

Correct me if I'm wrong @bighappyface, but I don't believe we have any of this available at this time and I'd imagine that would be a larger discussion to have down the road. Would it make sense to remove that rule from the linter for now and come back later?

As far as the wildcard selector goes, can you reference a specific child class instead (and also remove the nesting)? I don't really see the point of the wildcard here, but if you have a specific reason for it please let me know.

amberRrucker: For reusability. It acts as a psuedo mixin right now effecting I believe 4 or 5 patterns in multiple places. I would prefer to use it as a mixin for those classes than to define it again and again per element that needs flex applied to it but that's up to yalls interpretation of maintainable.

amberRrucker: Row-eq-height is also used pretty liberally in general content  as well.

WidgetsBurritos: @amberRrucker are you referring to the wildcard selector it the vendor prefixes?

SamuelRandall: @WidgetsBurritos the sass-linter error on that line is in regards to the nesting depth. Oddly enough it doesn't throw an error for the `*` operator, unless for some reason it is limited to only throwing one error per line.

SyntaxStacks:
Please Rebase

styleguide/_themes/derek/scss/_modifiers.scss
WidgetsBurritos: @SamuelRandall Well I think by using the wildcard, you're forced into that nesting anyway.

PR #117, opened 2016-06-06
rug and basement adjustments and zoolander width restriction

amberRrucker:


styleguide/_themes/derek/scss/components/_basement.scss
```diff
@@ -7,3 +7,11 @@
 .basement-copyright{
   color: $medium-gray;
 }
+.basement-region-list-americas {
```
WidgetsBurritos: replace with mobile-first approach so we only have to put `column-count: 2` as an exception, instead of overriding it .

styleguide/_themes/derek/scss/components/_rug.scss
```diff
@@ -1,10 +1,15 @@
 .rug-phone-numbers{
   margin-top:6px;
+  padding-left:5px;
+  a{
```
WidgetsBurritos: Should this `a` have a class so we can reference it directly, such as `.rug-phone-numbers-link`.

styleguide/derek/_partials/_basement.ejs
```diff
@@ -12,7 +12,7 @@
       </div>
       <div class="col-sm-3 col-sm-push-6">
         <p class="uppercase copyright basement-copyright">&reg; 2016 Rackspace US, Inc.</p>
-        <div id="teconsent" style="display:none;"></div>
+        <div id="teconsent""></div>
```
WidgetsBurritos: Not sure if `style="display: none;"` is needed, but if not we still have double quotes here that need to be fixed.

styleguide/_themes/derek/scss/components/_basement.scss
```diff
@@ -7,3 +7,11 @@
...
+  column-count: 2;
+}
+@media only screen and (max-width : 770px) {
```
WidgetsBurritos: Use bootstrap variable.

amberRrucker:
@WidgetsBurritos Thanks for the reminders! Pretty sure I got all those ironed out. @bighappyface is the naming here what you were looking for?

WidgetsBurritos:
:+1: on the rug portion.

I will test the basement portion in just a minute.

styleguide/derek/_partials/_basement.ejs
```diff
@@ -12,7 +12,7 @@
       </div>
       <div class="col-sm-3 col-sm-push-6">
         <p class="uppercase copyright basement-copyright">&reg; 2016 Rackspace US, Inc.</p>
-        <div id="teconsent" style="display:none;"></div>
+        <div id="teconsent" "style="display:none;"></div>
```
WidgetsBurritos: Extra quote here.

bighappyface:
+1

PR #119, opened 2016-06-09
RSWEB-6180: remove subnav cta visited focus active states

jhthompson:
https://jira.rax.io/browse/RSWEB-6180

Removes states which cause subnav cta to appear colors other than signup green.


bighappyface:
+1

PR #121, opened 2016-06-11
RSWEB-6385: default links to be underlined in content

writes:
https://jira.rax.io/browse/RSWEB-6385

- Underlines all links in content and removes the requirement of surrounding `<p>` tags. 
- Updates navigation and footer elements to maintain current styles.

### DIFF
--- a/styleguide/_themes/derek/scss/components/_footer.scss
+++ b/styleguide/_themes/derek/scss/components/_footer.scss
@@ -1,12 +1,19 @@
 .footer-menu-link {
   font-size: 1.2rem;
-  color: darken($light-gray, 15%);
+  color: $medium-light-gray;
+  text-decoration: none;
+  &:hover {
+    color: $medium-light-gray;
+    text-decoration: underline;
+  }
 }
 
 .footer-tab-link {
   font-size: 1.3rem;
   text-transform: uppercase;
   color: $white;
+  text-decoration: none;
+
   &:hover, &:focus {
     cursor: default;
     text-decoration: none;
--- a/styleguide/_themes/derek/scss/components/_rug.scss
+++ b/styleguide/_themes/derek/scss/components/_rug.scss
@@ -1,15 +1,50 @@
-.rug-phone-numbers{
-  margin-top:6px;
-}
-.rug-subhead{
-  font-size: 37px;
-  line-height: 24px;
-}
-.rug-contact-text{
-  margin-top: 15px;
-}
-.rug-contact-link{
-  text-decoration: none;
-  padding-left: 5px;
+.rug {
+  background-color: $blue-gray;
+  border-bottom: 1px solid $medium-light-gray;
 }
 
+.rug-chat-text {
+  padding: 15px 0 0 5px;
+}
+
+.rug-container {
+  padding-bottom: 25px;
+  padding-top: 25px;
+}
+
+.rug-email-text {
+  padding: 15px 0 0 5px;
+}
+
+.rug-head {
+  color: $white;
+  margin: 0;
+}
+
+.rug-icon {
+  color: $white;
+}
+
+.rug-link {
+  color: $white;
+  text-decoration: none;
+}
+
+.rug-link-hover {
+  &:hover, &:focus {
+    color: $white;
+    text-decoration: underline;
+  }
+}
+
+.rug-phone-text {
+  color: $white;
+  padding: 7px 0 0 5px;
+}
+
+.rug-subhead {
+  color: $white;
+  font-size: 37px;
+  line-height: 24px;
+  margin: 0;
+}
--- a/styleguide/_themes/global/scss/_swatch.scss
+++ b/styleguide/_themes/global/scss/_swatch.scss
@@ -7,6 +7,7 @@ $light-gray-white: #efefef;
 $bright-gray: #f2f5f5;
 $light-gray: #dedede;
 $medium-gray: #666;
+$medium-light-gray: #b8b8b8;
 $medium-dark-gray: #4c4c4c;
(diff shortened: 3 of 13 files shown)


writes:
@bighappyface I think this is ready to go. Let me know if you see any issues. Thanks!

styleguide/_themes/derek/scss/components/_footer.scss
```diff
@@ -1,12 +1,14 @@
 .footer-menu-link {
   font-size: 1.2rem;
   color: darken($light-gray, 15%);
+  text-decoration: none;
```
WidgetsBurritos: Hey we forgot to add this to Roger's original footer PR.  Can you add it here? 
```
  &:hover, &:focus {
    color: $white;
  }
```

Otherwise the mouseover links are blue and that looks weird on the dark background.

styleguide/_themes/derek/scss/components/_rug.scss
```diff
@@ -1,15 +1,18 @@
-.rug-phone-numbers{
-  margin-top:6px;
-}
-.rug-subhead{
-  font-size: 37px;
-  line-height: 24px;
+.rug-contact-link {
+  padding-left: 5px;
+  text-decoration: none;
```
WidgetsBurritos: Here too...
```
  &:hover, &:focus {
    color: $white;
  }
```

WidgetsBurritos: Alternatively, DP recommended doing:

```
.rug-contact-link-hover {
  &:hover, &:focus {
    color: $white;
    text-decoration: underline;
  }
}
```

and then adding the class to the HTML structure.

writes:
@WidgetsBurritos this has been updated with some of the suggestions we talked about. Let me know if you see any further issues.

Thanks!

styleguide/derek/_partials/_rug.ejs
```diff
@@ -3,28 +3,28 @@
...
     <div class="col-sm-3">
       <div class="row">
         <div class="col-md-2">
-          <i class="fa fa-mobile fa-4x white"></i>
+          <i class="fa fa-mobile fa-4x rug-icon"></i>
         </div>
         <div class="col-md-10 no-padding rug-contact-link rug-phone-numbers">
-          <span class="white"><a class="white rug-contact-link" href="tel:1-844-858-5963">1-844-858-5963</a> Sales</span><br/>
-          <span class="white"><a class="white rug-contact-link" href="tel:+18009614454">1-800-961-4454</a> Support</span>
+          <span class="rug-link"><a class="rug-link rug-link-hover" href="tel:1-844-858-5963">1-844-858-5963</a> Sales</span><br/>
```
WidgetsBurritos: Not sure if it makes sense to use the same class for the `<span>` and for the `<a>` tag here. Perhaps the span should be `rug-text` and the `a` should stay as `rug-link`.

styleguide/derek/_partials/_rug.ejs
```diff
@@ -33,10 +33,10 @@
       <a href="/information/contactus#form">
         <div class="row">
           <div class="col-md-3">
-            <i class="fa fa-envelope-o fa-3x white"></i>
+            <i class="fa fa-envelope-o fa-3x rug-icon"></i>
           </div>
           <div class="col-md-9 no-padding rug-contact-text">
-            <span class="white rug-contact-link">Email Us</span>
+            <span class="rug-link rug-link-hover rug-link rug-link-hover-hover">Email Us</span>
```
WidgetsBurritos: `rug-link-hover-hover`?

styleguide/_themes/global/scss/_swatch.scss
```diff
@@ -3,6 +3,7 @@ $images: "https://752f77aa107738c25d93-f083e9a6295a3f0714fa019ffdca65c3.ssl.cf1.
 
 //Colors
 $white: #fff;
+$footer-gray: #b8b8b8;
```
WidgetsBurritos: This is totally my opinion, and may be a question for @bighappyface, so don't necessarily act on it.  I'm just putting it out here for discussion. 

If we're gonna have region-specific colors, does it make sense to define them outside of that region (e.g. not in `_footer.scss` file in this case)? 

If so, that's fine, but perhaps _swatch.scss should be organized in such a manner to indicate that there is a master list of colors to choose from and then there are region specific colors, that reference those.  

So for example: 

```
$medium-light-gray: #b2b2b2; 
```

and then towards the end of the file: 
```
// Footer colors
$footer-color-link-hover: $medium-light-gray;
```

Otherwise, I think we may run the risk of redefining colors, or using inconsistent colors across various parts of the website. But once again, this is just my opinion so by all means provide additional feedback.

amberRrucker: @WidgetsBurritos The colors in swatches are global brand approved colors we're allowed to use. Any addition to swatches to be super specific to derek should be put into a derek/_swatches.scss instead of global. I dont love having to define new colors for a hover effect in the first place but we def dont want to go cluttering up a global file with region and hover specific colors. @bighappyface what do you think?

styleguide/_themes/derek/scss/components/_rug.scss
```diff
@@ -1,15 +1,35 @@
...
+  }
+}
+
+.rug-phone-numbers {
+  margin-top: 6px;
+  padding: 0;
 }
 
+.rug-subhead {
+  color: $white;
```
amberRrucker: @bighappyface do we want to define the color within each of these instead of just adding the class to the html or extending .white? Is there a benefit over other options?

WidgetsBurritos:
:+1:

PR #122, opened 2016-06-14
chore(package.json): Add "harp:compile" script

WidgetsBurritos:
Adds a harp:compile option to package.json to allow for zoolander compilation, with out adding to git repo.


writes:
+1

PR #125, opened 2016-06-15
last minute changes

amberRrucker:


styleguide/derek/_partials/_footer.ejs
```diff
@@ -1,5 +1,5 @@
 <!--[footernav]-->
-<div class="bg-blue-gray standard-padding-full footer">
+<div class="bg-blue-gray half-padding-full footer">
```
WidgetsBurritos: That's gonna require a change on the www side as well.  Can you apply that padding to `.footer` instead and then we can remove the padding classes in www in a future sprint?

bighappyface:
+1

styleguide/derek/_partials/_footer.ejs
WidgetsBurritos: Nevermind.  DP said to go ahead and do it in www.  I'll adjust.

PR #127, opened 2016-06-16
cornerPocket styling

amberRrucker:
Adding in the cornerPocket styling to zoolander. Also removing some unused swatches.


SyntaxStacks:
👍

PR #128, opened 2016-06-16
chore(lint): rsweb-6638 Sass Lint Class names

SyntaxStacks:
https://jira.rax.io/browse/RSWEB-6638

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -9,4 +9,6 @@ files:
 rules:
   no-url-protocols: 0
   class-name-format:
-    - convention: "^([a-z0-9]+([A-Z][a-z0-9]+)*)(-{1}([a-z0-9]+([A-Z][a-z0-9]+)*)){0,2}$"
+    - 1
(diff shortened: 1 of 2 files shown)


bighappyface:
With the sass lint PR merged how are there 1675 problems failing this build?

SyntaxStacks:
I removed the changes to the sass files and brought in sass-lint.

The builds have been failing since travis was integrated so there's not too much harm in that and after our adoption of maintainable css, I felt that it would be better to refactor the code to reflect that while correcting the lint issues. one thing that will have to be addressed is the max warning flag on the lint job. I have removed that flag in this PR to allow future bumps and commits.

SyntaxStacks:
we can add `--max-warnings 0` back to the `sass-lint` npm script when we're ready to enforce sass-lint warnings as build failures

bighappyface:
+1 on the convention update

Example of the regex in action http://refiddle.com/dzxm

PR #130, opened 2016-06-20
RSWEB-6655: o365 sass fixes

writes:
https://jira.rax.io/browse/RSWEB-6655
Fixes the majority of lint errors without changing styles.

4 errors remain. 

2 - ID selectors being used
1 - nesting greater than 3 
1 - class should match regular expression


WidgetsBurritos:
:+1:

PR #141, opened 2016-06-21
RSWEB-6655: global-scss-global sass fixes

SamuelRandall:
https://jira.rax.io/browse/RSWEB-6655

Leading underscore errors persist regarding the names of the imported files.


WidgetsBurritos:
:+1: although this will be a conflict with #144 so depending on which is merged first, one of these will require a rebase.

PR #144, opened 2016-06-21
RSWEB-6673 Convert zoolander ceiling to use maintainable CSS.

WidgetsBurritos:
Jira: https://jira.rax.io/browse/RSWEB-6673

This PR converts the ceiling into maintainable CSS.

### DIFF
--- a/styleguide/_themes/global/scss/global.scss
+++ b/styleguide/_themes/global/scss/global.scss
@@ -5,3 +5,4 @@
 @import '_fonts';
 @import '_forms';
 @import '_buttons';
+@import '_bootstrap_overrides';
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -5,6 +5,7 @@ files:
     - 'styleguide/_sass/_syntax-highlighting.scss'
     - 'styleguide/_themes/derek/scss/components/_rsweb-font-style.scss'
     - 'styleguide/_themes/global/scss/_bootstrap-variables.scss'
+    - 'styleguide/_themes/global/scss/_bootstrap-overrides.scss'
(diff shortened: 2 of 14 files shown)


styleguide/_themes/global/scss/global.scss
```diff
@@ -5,3 +5,4 @@
 @import "_fonts";
 @import "_forms";
 @import "_buttons";
+@import '_bootstrap_overrides';
```
WidgetsBurritos: Note: To comply with lint standards (and also play nicely with #141), I have opted to use single quotes in just this one instance.  Since #141 fixes the rest of the file I opted to leave that alone.

bighappyface:
@WidgetsBurritos the commit here is under your SSO

styleguide/_themes/global/scss/global.scss
```diff
@@ -5,3 +5,4 @@
 @import '_fonts';
 @import '_forms';
 @import '_buttons';
+@import '_bootstrap_overrides';
```
bighappyface: I am in favor of a PR that removes the unnecessary underscores on these and other files. In Harp underscores are meant to be excluded but the `_themes` directory excludes everything within it, and the exclusions do not influence Sass imports. We could avoid the single/double quote thing altogether without underscores.

WidgetsBurritos: I think that may already be in the works.  Should we just wait till then or do it now?

bighappyface:
+1

styleguide/_themes/global/scss/global.scss
bighappyface: Another PR that is just a bunch of filename changes (no line changes) is my preference. This is good to go.

PR #149, opened 2016-06-21
RSWEB-6655: basement sass fixes

writes:
https://jira.rax.io/browse/RSWEB-6655

PR #151, opened 2016-06-21
RSWEB-6655: Fix basement flag bug

WidgetsBurritos:
PR #149 introduced a bug in the basement that I missed the first time around. 

Otherwise the basement flags won't show up.


SyntaxStacks:
👍

bighappyface:
+1

PR #159, opened 2016-06-22
chore(main): Added color in place of missing variable

SyntaxStacks:


SyntaxStacks:
Had to up the error threshold as this brought in another linting error

WidgetsBurritos:
👍

PR #162, opened 2016-06-28
chore(basement): Used Mixins for basement

SyntaxStacks:
https://jira.rax.io/browse/RSWEB-6706

### DIFF
--- a/styleguide/_themes/derek/scss/globalElements/basement.scss
+++ b/styleguide/_themes/derek/scss/globalElements/basement.scss
@@ -7,12 +7,17 @@
 
 // Basement Classes
 .basement {
-  padding: 10px 15px;
+  @include make-container;
+  padding: 10px 0;
 }
 
 .basement-copyright {
   color: $gray-base;
   text-align: right;
+}
+
+.basement-row {
+  @extend %row;
 }
 
 .basement-flag-icon {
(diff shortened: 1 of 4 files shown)


WidgetsBurritos:
👍

SyntaxStacks:
https://github.rackspace.com/rswebteam/www/pull/1191

styleguide/derek/_partials/_basement.ejs
```diff
@@ -1,17 +1,15 @@
 <!--[basement]-->
 <div class="bg-dark-gray">
-  <div class="container basement">
+  <div class="basement">
     <div class="row">
```
WidgetsBurritos: lets change `row` class to `basement-row` and then use the `make-row()` mixin in css.

styleguide/_themes/global/scss/_grid.scss
```diff
@@ -0,0 +1,22 @@
+@mixin make-container-fluid {
+  @include container-fixed;
+}
+
+@mixin make-container {
```
WidgetsBurritos: This makes me happy.

WidgetsBurritos:
👍 for the revisions.

styleguide/derek/_partials/_basement.ejs
SyntaxStacks: I created a `row` class that simply includes the mixin `make-row`
I'd rather not have duplicate css for rows per module but if we would like to take that approach I'd prefer we have a row placeholder class so that the sass will output multiple classes with the same row styling in one place.

styleguide/_themes/global/scss/_grid.scss
```diff
@@ -0,0 +1,22 @@
...
+  }
+  @media (min-width: $screen-md-min) {
+    width: $container-md;
+  }
+  @media (min-width: $screen-lg-min) {
+    width: $container-lg;
+  }
+}
+
+.row {
```
bighappyface: `.row` already exists in Bootstrap in this exact fashion:

https://github.com/twbs/bootstrap/blob/10606a73bcbfc6c723d60e2a1069c921afff0d2a/less/grid.less#L39-L41
https://github.com/twbs/bootstrap/blob/daad359f016949725f4ba39645e3f9ade7dda246/dist/css/bootstrap.css#L1612

In the spirit of MaintainableCSS having the properties duplicated is not a bad thing and provides better defense against regression; however, that is somewhat lessened when mixins are involved as anything including it could be affected if there were a change.

If we are dumping the columns in favor of using mixins then dumping a row in favor of a mixin seems fine too, no? I agree that a placeholder class would be well-suited so we get the best of both worlds.

styleguide/_themes/derek/scss/components/_basement.scss
```diff
@@ -7,12 +7,12 @@
...
-  padding: 10px 15px;
+  @include make-container;
+  padding: 10px 0;
 }
 
-.basement-copyright {
-  color: $medium-gray;
-  text-align: right;
+.basement-row {
+  @extend %row;
```
WidgetsBurritos: This should probably just be `@include make-row;` not `@extend %row;` as that's all it's doing anyway. Seems like we're adding an extra step here by using and extend

styleguide/_themes/global/scss/_grid.scss
```diff
@@ -0,0 +1,22 @@
+@mixin make-container-fluid {
```
WidgetsBurritos: So grid.scss should exist now (in lieu of this _grid.scss) file (see: https://github.com/rackerlabs/zoolander/blob/master/styleguide/_themes/global/scss/grid.scss) 

Most of this functionality should match that. The only think I don't have is your `%row` placeholder, although I still think we should just skip that and use `@include make-row;` in `.basement-row` instead, as that's how we're doing it elsewhere throughout the site.

styleguide/_themes/derek/scss/globalElements/basement.scss
```diff
@@ -25,61 +30,78 @@
...
 }
 
-@include basement-flag(argentina, 0, -280px);
+.basement-copyright {
+  @include make-sm-column(3);
+  @include make-sm-column-offset(6);
+}
+
+.basement-copyright-text {
+  color: $medium-gray;
```
WidgetsBurritos: `$medium-gray` is no longer a valid color.  Consult the variables here: https://github.com/rackerlabs/zoolander/blob/master/styleguide/_themes/global/scss/swatch.scss

styleguide/_themes/derek/scss/globalElements/basement.scss
```diff
@@ -7,14 +7,19 @@
...
+  padding: 10px 0;
 }
 
 .basement-copyright {
   color: $gray-base;
   text-align: right;
 }
 
+.basement-row {
+  @extend %row;
```
WidgetsBurritos: To reiterate what I said elsewhere, I think for consistency purposes it makes sense to just do `@include make-row;` here instead, since that's all that extension is doing anyway.

PR #163, opened 2016-06-28
Main Navigation

amberRrucker:
Open for code review. In progress on tertiary subnav breaking right page container. Everything else is good to review.
https://jira.rax.io/browse/RSWEB-6226


styleguide/js/zoolander.js
```diff
@@ -1,3 +1,26 @@
+jQuery(document).ready(function(){
```
WidgetsBurritos: This will be the third `jQuery(document).ready()` function in `zoolander.js`.  Since the file doesn't really do all that much, it seems it would make more sense to just combine them into a single `jQuery(document).ready()` function.   

Additionally, make sure the function declaration has the `$` variable declared, if you plan on using `$` instead of `jQuery`. 
```
jQuery(document).ready(function ($) {
```

After doing this, any nested `jQuery()` functions could probably just be replaced with `$` to make code a little more succinct.

styleguide/js/zoolander.js
```diff
@@ -1,3 +1,26 @@
...
+    $(".navbar-searchButton").css('opacity', 1);
+  })
+  jQuery (".navbar-hamburger").click(function(){
+    $(".navbar-hamburger").toggleClass("collapsed");
+    $(".navbar-search").slideUp();
+    $(".navbar-activeArrow").fadeOut();
+  })
+  jQuery (".navbar-tertiary-dropDownTrigger").hover(function(e){
+    var current=$(this).find('.navbar-tertiary-dropDownMenu');
+    console.log(current);
```
WidgetsBurritos: Remove this console.log()

styleguide/_themes/derek/scss/components/subnav.scss
```diff
@@ -0,0 +1,87 @@
...
+        border-bottom: solid transparent 4px;
+      }
+
+    }
+
+  }
+}
+
+//subnav
+@media only screen and (max-width: $screen-md) {
```
WidgetsBurritos: I know this is a carry over from existing code, but this max-width should probably be set to `$screen-sm-max` instead of `$screen-md`.

styleguide/_themes/derek/scss/components/subnav.scss
```diff
@@ -0,0 +1,87 @@
...
+  .subnav li {
+    font-size: .86em;
+    margin: 0;
+    padding: 7px 0 6px;
+  }
+}
+
+@media only screen and (max-width: $screen-sm) {
```
WidgetsBurritos: same here.

WidgetsBurritos:
This will need to be ran through the sass-lint process again.  This PR bumped the error count from 127 to 1002. I'm assuming some of which is related to legacy code from prior to the sass linting tasks, but those will need to be addressed.

styleguide/_themes/derek/scss/components/mainnav.scss
```diff
@@ -0,0 +1,371 @@
+//main nav
+.main-nav {
+  background: $dark-gray;
+  border: 0;
+  border-radius: 0;
+  margin: 0;
+
+  .navbar-menuContainer {
```
WidgetsBurritos: Any nesting and other code that becomes necessary due to overriding bootstrap, should go into the `styleguide/_themes/global/scss/bootstrap_overrides.scss` file.

That file is exempt from sass linting so we can get away with doing that kind of stuff there. That said, that file should be used sparingly, and only for properties that must be that specific to override. So if there are any properties you don't need that level of specificity for, place them in a top-level class instead. 

Additionally, if you make use of the bootstrap override file in a component, add a comment at the top of that file indicating some property overrides exist in the separate file to override bootstrap-specific styles, to assist with future maintainability.

styleguide/_themes/derek/scss/components/mainnav.scss
```diff
@@ -0,0 +1,371 @@
...
+        background-color: transparent;
+      }
+
+    }
+  }
+}
+
+.navbar-dropDown-trigger {
+
+  &::after {
```
WidgetsBurritos: I'm assuming this should be `&:after` instead?

WidgetsBurritos: Actually disregard that comment. I'm wrong. I wasn't paying attention.

styleguide/_themes/derek/scss/components/mainnav.scss
```diff
@@ -0,0 +1,371 @@
...
+}
+
+.navbar-footer {
+  bottom: 0;
+  position: fixed;
+  width: 100%;
+}
+
+.navbar-footerLink {
+  @extend h6;
```
WidgetsBurritos: This will obviously get picked up by the sass linter, but all @extends should use placeholders instead of classes, use a mixin instead, or just place the properties directly into the selector.

styleguide/_themes/derek/scss/components/mainnav.scss
```diff
@@ -0,0 +1,371 @@
...
+  float: right;
+  margin: 0;
+  opacity: .7;
+  padding: 6.5px;
+  width: 20%;
+}
+
+
+//main nav show dropdowns on hover on desktop
+@media only screen and (min-width: $screen-md) {
```
WidgetsBurritos: `$screen-md` (and all other size screen vars) are deprecated as of bootstrap 3.0.1. We should start using `$screen-md-min` and `$screen-md-max` instead. 

`max-width:` media queries will always be used with `$screen-*-max`
`min-width:` media queries will always be used with `$screen-*-min`

That makes the code much more readable and will help prevent one-off errors right at breakpoints.

styleguide/js/zoolander.js
amberRrucker: I mentioned to DP I wrote this section sort of long form just to be super separate about the js until it was approved, made sure it was functioning properly and good to go. If its all working as everyone expects I can slim this section down pretty significantly.

styleguide/_themes/derek/scss/components/mainnav.scss
amberRrucker: I wanted to make sure that for code review purposes that everything was mostly all in one place and when I wrote this there was no bootstrap_overrides file. I will be sure to move all of that over once I address the rest of the concerns in here. Thank ya sir. I'll get with you on how you would prefer that file organized.

styleguide/_themes/derek/scss/components/mainnav.scss
amberRrucker: I will go ahead and just add it to the selector. Thank ya.

styleguide/_themes/derek/scss/components/subnav.scss
amberRrucker: My sass-lint is giving me nowhere near that many errors and theyre all on the nesting (which i mentioned in a comment above), we will have to sync up so I can see what youre seeing.

styleguide/js/zoolander.js
WidgetsBurritos: Okay that's fine.

styleguide/_themes/derek/scss/components/subnav.scss
WidgetsBurritos: Okay I can show you in the morning.

styleguide/_themes/derek/scss/components/mainnav.scss
WidgetsBurritos: Yeah no problem.

styleguide/_themes/derek/scss/components/subnav.scss
WidgetsBurritos: Just FYI, you can see all sass lint failures on the Travis job (which actually has a slightly different number than on my local machine.  Not sure why)
https://travis-ci.org/rackerlabs/zoolander/jobs/140928430

Compare that with the Travis job for the most recently merged PR: 
https://travis-ci.org/rackerlabs/zoolander/jobs/139517733

styleguide/_themes/derek/scss/components/mainnav.scss
WidgetsBurritos: Just making a note here for myself and other code reviewers that this nesting was deemed necessary due to bootstrap stuff, and do in fact pass the sass-lint process, so there is no need to relocate it after all.

.gitignore
```diff
@@ -6,3 +6,4 @@ bower_components
 Gemfile.lock
 styleguide/css/*.css
 backstop_data/bitmaps_test
+npm-debug.log
```
bighappyface: This is best done in a [global .gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore). Would anyone EVER want to add that to a repo?

WidgetsBurritos: @bighappyface Does it hurt having it in both?

styleguide/_themes/derek/scss/components/mainnav.scss
```diff
@@ -0,0 +1,354 @@
...
+  margin: 0;
+
+  .navbar-menuContainer {
+    border: 0;
+    max-height: none;
+    padding: 0;
+  }
+}
+
+.navbar-nav {
```
bighappyface: I don't quite understand the need for this wrapper. Classes like `.navbar-topLink` should be unique enough to stand on their own.

amberRrucker: It isnt specific enough unfortunately. Bootstrap has things like .navbar > ul > li > a and navbar-topLink doesnt work outside of the wrapper.

.gitignore
SyntaxStacks: The same could be said about most of the files in the `.gitignore` file. It's better to not allow it into the app via the repo's `.gitignore` than assuming every contributor will have added this to the their global gitignore file.

styleguide/_themes/derek/scss/components/mainnav.scss
bighappyface: In this case I suppose it could be worse.

With the sheer amount of styling and overrides happening in this PR I question the need for using bootstrap's `.navbar` classes altogether.

amberRrucker: It's nice to be able to use the navbars baked in functionality. The styling is going to have to happen anyways. I didnt think there were that many overrides really. Up to you but I didnt think it was too much by the way of overrides.

bighappyface: I am not asking to rework it, but i do think it could be a cleaner implementation. Like I said, it could be worse.

WidgetsBurritos: @bighappyface Unfortunately navbar behavior isn't available as a mixin 
(see: https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_navbar.scss)

We could theoretically create a mixin for our own personal use to emulate some of that, but that may be better suited for another PR.  @SyntaxStacks is already doing a little bit of that in his basement PR for containers, so we could maybe tie into that down the road. 

Thoughts?

.gitignore
bighappyface: Bikeshedding...

My comment was a recommendation and I have to make this same recommendation all of the time to developers all over the world.

styleguide/_themes/derek/scss/components/mainnav.scss
bighappyface: I will take this iteration and we can litigate it further at a later time.

WidgetsBurritos:
👍

PR #164, opened 2016-06-29
RSWEB-6329: Fixes harp:compile bug

WidgetsBurritos:
This PR is a quick fix for bug introduced by #163.


amberRrucker:
+1 good catch! missed that one.

PR #165, opened 2016-06-29
RSWEB-6329: Convert tertiary menu triggers from anchors to spans.

WidgetsBurritos:
This PR converts the tertiary triggers to `<span>` tags from `<a>` tags.  This was done in #163 for top-level nav, but got overlooked in the tertiary triggers.


amberRrucker:
+1

SyntaxStacks:
👍

PR #167, opened 2016-06-30
making minor adjustments to main nav

amberRrucker:
Couple issues introduced by mixing spans and anchor tags and bootstraps subsequent freakout.

### DIFF
--- a/styleguide/_themes/derek/scss/components/mainnav.scss
+++ b/styleguide/_themes/derek/scss/components/mainnav.scss
@@ -17,6 +17,7 @@
   .navbar-topLink {
     color: $white;
     display: block;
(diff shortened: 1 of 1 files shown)


styleguide/_themes/derek/scss/components/mainnav.scss
```diff
@@ -268,6 +270,10 @@
 
     .navbar-topItem {
       font-size: 1.07rem;
+
+      .navbar-dropDownLink {
```
WidgetsBurritos: As much as I hate this nested like this, I've confirmed its necessary due to weird bootstrap specificity.

WidgetsBurritos:
👍

PR #173, opened 2016-07-05
feat(ticker): rsweb-6705 Openstack Ticker

SyntaxStacks:
https://jira.rax.io/browse/RSWEB-6705
screen shot 2016-07-05 at 12 27 07 pm

### DIFF
--- a/styleguide/_themes/derek/scss/components/ticker.scss
+++ b/styleguide/_themes/derek/scss/components/ticker.scss
@@ -1,0 +1,95 @@
+@import '../../../global/scss/swatch';
+@import '../../../global/scss/fonts';
+@import '../../../global/scss/bootstrap_variables';
+@import '../../../../../node_modules/bootstrap-sass/assets/stylesheets/bootstrap/mixins/clearfix';
+@import '../../../../../node_modules/bootstrap-sass/assets/stylesheets/bootstrap/mixins/grid';
+
+$ticker-span-bg: rgba(200, 0, 0, .5);
+
+.ticker {
+  background-color: $rackspace-red;
+  width: 100%;
+
+  a {
+    text-decoration: none;
+  }
+}
+
+.ticker-container {
+  @include container-fixed;
+  max-width: 960px;
+
+  @media (min-width: $screen-sm-min) {
+    width: $container-sm;
+  }
+  @media (min-width: $screen-md-min) {
+    width: $container-md;
+  }
+  @media (min-width: $screen-lg-min) {
+    width: $container-lg;
+  }
+}
+
+.ticker-row {
+  @include make-row;
+}
+
+.ticker-label {
+  @include make-md-column(4);
+  color: $white;
+  font-family: $subhead;
+  padding: 30px 0;
+  text-align: center;
+  text-transform: uppercase;
+}
+
+.ticker-countdown-column {
+  @include make-md-column(4);
+  padding: 0;
+}
+
+.ticker-countdown {
+  background-color: $white;
+  color: $black;
+  font-family: $subhead;
+  font-size: 29px;
+  font-weight: 700;
+  letter-spacing: 3px;
+  margin: 30px;
+  padding: 15px;
+  text-align: center;
+}
+
+.ticker-image {
+  @include make-md-column(4);
+  background-image: url('https://752f77aa107738c25d93-f083e9a6295a3f0714fa019ffdca65c3.ssl.cf1.rackcdn.com/swirlBG.svg');
+  background-position: bottom;
+  background-repeat: no-repeat;
+  background-size: contain;
+  min-height: 113px;
+  padding: 0;
+}
+
+@media only screen and (max-width: $screen-sm-max) {
+  .ticker-image {
+    display: none;
+  }
+
+  .ticker-label {
+    padding-bottom: 0;
+  }
+
+  .ticker-row {
+    background-image: url('https://752f77aa107738c25d93-f083e9a6295a3f0714fa019ffdca65c3.ssl.cf1.rackcdn.com/swirlBG.svg');
+    background-position: bottom;
+    background-repeat: no-repeat;
+    background-size: contain;
+  }
+
+  .ticker-span-bg {
+    background-color: $ticker-span-bg;
+    font-size: 24px;
+    line-height: 30px;
+    padding: 2px;
+  }
+}
--- a/styleguide/derek/_partials/ticker.ejs
+++ b/styleguide/derek/_partials/ticker.ejs
@@ -1,0 +1,16 @@
+<section class="ticker">
+  <a href="https://www.rackspace.com/cloud/lp/openstack-expertise">
+    <section class="ticker-container">
+      <div class="ticker-row">
+        <div class="ticker-label" id="ticker-label">
+          <span class="ticker-span-bg">Nearing 1 billion <br> managed server hours</span>
+        </div>
+        <div class="ticker-countdown-column">
+          <div class="ticker-countdown" id="ticker-countdown">998,897,493</div>
+        </div>
+        <div class="ticker-image">
+        </div>
+      </div>
+    </section>
+  </a>
+</section>
--- a/styleguide/derek/examples/landing-page.ejs
+++ b/styleguide/derek/examples/landing-page.ejs
@@ -8,7 +8,7 @@
             <p><a class="btn btn-primary btn-lg" href="#" role="button">Learn more »</a></p>
         </div>
     </div>
-
+    <%- partial("../_partials/ticker") %>
     <div class="container">
         <div class="col-sm-4">
             <h3 class="text-center">Text</h3>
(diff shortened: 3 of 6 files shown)


WidgetsBurritos:
Code: +1

WidgetsBurritos:
Actually I retract my previous +1.

Is the text supposed to line up with the rug or footer text?

styleguide/derek/_partials/ticker.ejs
```diff
@@ -0,0 +1,12 @@
...
+    <div class="row">
+      <div class="ticker-label">Count to 1 Billion Node Hours!</div>
+      <div class="ticker-countdown-column">
+        <div class="ticker-countdown" id="countdown">998,897,493</div>
+      </div>
+      <div class="ticker-image">
+      </div>
+    </div>
+  </div>
+</div>
```
WidgetsBurritos: You have open `<section>` tags, but closing `<div>` tags.

styleguide/derek/_partials/ticker.ejs
```diff
@@ -0,0 +1,12 @@
+<section class="ticker">
+  <section class="container">
+    <div class="row">
```
WidgetsBurritos: Didn't you create a `make-container()` mixin? Then we could have a ticker-container class instead of standard container class for maintainable css purposes. 

Also, use ticker-row instead of row and then do `make-row();`

styleguide/derek/_partials/ticker.ejs
```diff
@@ -0,0 +1,12 @@
+<section class="ticker">
+  <section class="container">
```
WidgetsBurritos: Should only the outer tag be a section, and the inner container be a div?

styleguide/_themes/derek/scss/components/ticker.scss
```diff
@@ -0,0 +1,38 @@
+.ticker {
+  background-color: $rackspace-red;
+  width: 100%;
```
WidgetsBurritos: Is `width: 100%;` entirely necessary? 
If this is happening out outside of a container, isn't already set at full width? (Just a question.  I haven't tested it).

styleguide/_themes/derek/scss/components/ticker.scss
```diff
@@ -0,0 +1,38 @@
...
+  text-align: center;
+}
+
+.ticker-image {
+  @include make-md-column(4);
+  background-image: url('https://752f77aa107738c25d93-f083e9a6295a3f0714fa019ffdca65c3.ssl.cf1.rackcdn.com/swirlBG.svg');
+  background-size: cover;
+  min-height: 110px;
+
+  @media only screen and (max-width: $screen-md-max) {
```
WidgetsBurritos: After talking with @bighappyface, it's probably better to place the media query outside of the `.ticker-image` instead of inside.  In the future we may look at implementing separate mobile css files, and if so, it would be easier to parse them out if they weren't nested inside.

styleguide/_themes/derek/scss/components/ticker.scss
```diff
@@ -0,0 +1,38 @@
...
+}
+
+.ticker-countdown-column {
+  @include make-md-column(4);
+}
+
+.ticker-countdown {
+  background-color: $white;
+  font-family: $subhead;
+  font-size: xx-large;
```
WidgetsBurritos: We should probably stick with a standard unit for font sizes instead of generic sizes like `xx-large`

styleguide/derek/_partials/ticker.ejs
WidgetsBurritos: Oh your `make-container()` mixin hasn't been merged in yet.  So I'd say we either just stick with `container` and add `ticker-row` or add that mixin to this PR instead of the basement PR and then just rebase that one later (as it needs a rebase anyway).

styleguide/_themes/derek/scss/components/ticker.scss
SyntaxStacks: Yes, this widget will be on external properties as well

WidgetsBurritos:
image

It might just be me but the way the text breaks looks a little strange on `xs` and `lg` screens.  Can we add something like `<br class="visible-xs visible-lg" />` after the word `hours` so `Managing OpenStack` is on its own line?

WidgetsBurritos:
image

That white text on white swirls is kinda hard to read.

WidgetsBurritos:
image

On `md` view the swirls are cutting off on the edges.

styleguide/_themes/derek/scss/components/ticker.scss
```diff
@@ -0,0 +1,88 @@
+@import '../../../global/scss/swatch';
+@import '../../../global/scss/fonts';
+@import '../../../global/scss/bootstrap_variables';
+@import '../../../../../node_modules/bootstrap-sass/assets/stylesheets/bootstrap/mixins/clearfix';
+@import '../../../../../node_modules/bootstrap-sass/assets/stylesheets/bootstrap/mixins/grid';
```
WidgetsBurritos: If we're gonna have an externally accessible css file, I'm not sure placing it inside the derek components directory makes sense.  Should this be in a `/styleguide/css/ticker.scss` file instead and then just have derek.scss include that instead?   (I'm not sure that makes sense or not, but having this inside derek specifically seems a little odd to me).

styleguide/derek/examples/landing-page.ejs
```diff
@@ -8,7 +8,7 @@
             <p><a class="btn btn-primary btn-lg" href="#" role="button">Learn more »</a></p>
         </div>
     </div>
-
+    <div id="countdown" style="margin-top: -30px"> </div>
```
WidgetsBurritos: Can this inline style be shifted to a css file?

styleguide/derek/examples/landing-page.ejs
```diff
@@ -33,3 +33,9 @@
         </div>
     </div>
 </div>
+<script src="https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/test/rackspace-js-lib-ticker.js?mo"></script>
```
WidgetsBurritos: Is this `test` folder on the cdn going to be the final location of this javascript file, or is this the URL someone else gave to you? I'm just curious.

WidgetsBurritos:
image

I see what you did there with the slightly transparent red background to mute the white in the background.  I think it's better but it does appear to have 1px of the original white sticking out above it.  I think if you're gonna do this, then that shouldn't happen.  Thoughts?

WidgetsBurritos:
Btw, you can barely tell on that screenshot its there but I can definitely see it in my browser.

styleguide/_themes/derek/scss/components/ticker.scss
```diff
@@ -0,0 +1,90 @@
...
+  @include make-row;
+}
+
+.ticker-label {
+  @include make-md-column(4);
+  color: $white;
+  font-family: $subhead;
+  padding: 30px 0;
+  text-align: center;
+  text-shadow: 1px 2px $black;
```
amberRrucker: lets nix this shadow and then im good with it!

styleguide/_themes/derek/scss/components/ticker.scss
```diff
@@ -0,0 +1,90 @@
...
+  text-transform: uppercase;
+}
+
+.ticker-countdown-column {
+  @include make-md-column(4);
+  padding: 0;
+}
+
+.ticker-countdown {
+  background-color: $white;
```
amberRrucker: are we no longer doing @extend .bg-white and what not? @WidgetsBurritos i havent heard anything about no longer doing that. have you?

WidgetsBurritos: I think we're trying to avoid @extends unless absolutely necessary.

SyntaxStacks: no extending classes

amberRrucker: okay cool then im fine with it. is that a maintainable thing or what not? wouldnt we want to reuse simple extendable code?

WidgetsBurritos: It's more to prevent issues with media queries in the future.

SyntaxStacks: in sass you should only extend placeholder, I learned that after I taught you to do this :/

amberRrucker: should we back background colors mixins then?

amberRrucker:
I see the tiny 1px white there now @WidgetsBurritos. So much detail haha. Lets increase the padding to 2px. then +1 fa shizz.

styleguide/_themes/derek/scss/components/ticker.scss
WidgetsBurritos: I just don't think its really that necessary.  It's just as easy to change a variable value as it is to change a mixin content.

WidgetsBurritos:
Code +1

amberRrucker:
+1 - looks the same. get it outta here.

PR #174, opened 2016-07-05
bug(styleguide):add padding to features icon

amberRrucker:
https://jira.rax.io/browse/RSWEB-6176
Adding in product icon padding. Went ahead and updated some naming to be more maintainable friendly. Didn't do a full refactor but its closer to the goal.

Will need an adjustment to the pattern in www for class names. If everyone is good with that then I'll go ahead and open a www PR.

### DIFF
--- a/styleguide/_themes/derek/scss/patterns/grid/productFeatures.scss
+++ b/styleguide/_themes/derek/scss/patterns/grid/productFeatures.scss
@@ -1,7 +1,7 @@
-.features {
+.features-container {
   display: flex;
 
-  &.three-up {
+  &.features-threeUp {
     display: block;
   }
 
@@ -11,17 +11,18 @@
     }
   }
 
-  .desc {
+  .features-desc {
     background-color: $white;
     margin: 0 15px 0 0;
     overflow: auto;
   }
 
-  .icon {
+  .features-icon {
     align-items: center;
     background-color: $darkteal;
     display: flex;
     justify-content: center;
+    padding: 10px;
   }
 
   .rsweb {
@@ -30,11 +31,11 @@
 }
 
 @media only screen and (max-width: $screen-sm) {
-  .features {
+  .features-container {
     display: block;
     margin-top: 10px;
 
-    .desc {
+    .features-desc {
       margin: 0;
     }
   }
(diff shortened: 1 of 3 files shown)


styleguide/_themes/derek/scss/patterns/grid/productFeatures.scss
```diff
@@ -30,11 +31,11 @@
 }
 
 @media only screen and (max-width: $screen-sm) {
-  .features {
+  .features-container {
     display: block;
     margin-top: 10px;
 
-    .desc {
+    .features-desc {
```
WidgetsBurritos: Can we add the `make-md-column()` here and get rid of the column classes while we're at it?

WidgetsBurritos: Correction: that would go outside of this media query, not inside

WidgetsBurritos: On second thought since this is a bug fix and not a refactor task, maybe we shouldn't risk introducing new bugs into the system with a lot of refactoring.  @amberRrucker @bighappyface thoughts?

amberRrucker: Yeah it's a bug fix and I snuck in some naming changes. I don't want to go full refactor on a super quick task for right now if that's fine with everyone.

WidgetsBurritos:
👍 on code changes. @runrog can you give this a design review?

runrog:
@WidgetsBurritos @amberRrucker adding the padding definitely looks better, the only thing I would suggest is to lower the font size of the icon on mobile at least. To something like 3 or 4em. They're kind of taking up a lot of real estate. 

5em
5em

3em
3em

amberRrucker:
@runrog I'll get with the ECMs. There were conversations around maintaining a larger size as a more defined "header" style area.

xinbin:
Hey there all, does this need to be altered in some magical way to get the checks to pass? Also this is on Hold. can it be merged?

amberRrucker:
@xinbin it can be merged as long as it isnt bumped until the corresponding PR in www goes out.

PR #175, opened 2016-07-07
chore(contrib): add contributors file

SyntaxStacks:


WidgetsBurritos:
@SyntaxStacks are we just setting this up as a kind of blank slate now, or should we start adding stuff to this file?   I know @amberRrucker was working on Zoolander docs, so not sure if the contributors file was included in that scope.

SyntaxStacks:
This should have been there from the start and while not included in the scope of @amberRrucker 's current work, it could be easily incorporated via an import in harp. I'll leave that up to her.

amberRrucker:
I'm down with having it in its own PR. My documentation task will likely be split into 4 or 5 PRs anyways because of the huge amounts of changes. I think having more people contribute to the effort the better. If anyone feels like adding anything to that file then go to town. My task is mostly documentation for styleguide. icons, available colors, font usage, ect. There will need to be more discussion and contribution around playbooks, processes and contributors stuff.

WidgetsBurritos:
Okay, we may want to at least link to bootstrap and maintainable for now.

amberRrucker:
+1

PR #177, opened 2016-07-07
RSWEB-6703: ias hover refactor

SamuelRandall:
https://jira.rax.io/browse/RSWEB-6703

### DIFF
--- a/styleguide/_themes/derek/scss/patterns/grid/ias.scss
+++ b/styleguide/_themes/derek/scss/patterns/grid/ias.scss
@@ -1,0 +1,115 @@
+.ias-container {
+  @include container-fixed;
+  max-width: 960px;
+  padding: $standard-padding;
+
+  @media (min-width: $screen-sm-min) {
+    width: $container-sm;
+  }
+  @media (min-width: $screen-md-min) {
+    width: $container-md;
+  }
+  @media (min-width: $screen-lg-min) {
+    width: $container-lg;
+  }
+}
+
+.ias-row {
+  @include make-row;
+}
+
+@media only screen and (min-width: $screen-sm-min) {
+  .ias-row {
+    @include flex-box;
+  }
+}
+
+.ias-col-full {
+  @include make-md-column(12);
+  @include flex-box;
+}
+
+.ias-col-half {
+  @include make-sm-column(6);
+  @include flex-box;
+}
+
+.ias-col-third {
+  @include make-xs-column(12);
+  @include make-sm-column(4);
+  @include flex-box;
+}
+
+.ias5-col-push {
+  @include make-sm-column-push(2);
+}
+
+.ias5-col {
+  @include make-sm-column(4);
+  @include flex-box;
+}
+
+.ias-title {
+  text-align: center;
+}
+
+.iasWidget {
+  padding: 20px 10px 0;
+  width: $full-width;
+}
+
+.iasWidget-body {
+  color: $darkest-gray;
+  font-family: $copy;
+  font-weight: 300;
+  line-height: 1.5em;
+  margin: 0 0 1em;
+  text-align: center;
+}
+
+.iasWidget-col-full {
+  @include make-md-column(12);
+}
+
+.iasWidget-cta {
+  color: $rackspace-red;
+  font-family: $copy;
+  font-weight: 600;
+  text-align: center;
+}
+
+.iasWidget-header {
+  color: $darkest-gray;
+  font-family: $heading;
+  font-size: 1.25em;
+  font-weight: 400;
+  line-height: 1.25em;
+  margin: 10px 0;
+  text-align: center;
+}
+
+.iasWidget-hover {
+  text-decoration: none;
+
+  &:hover {
+    background-color: $light-gray;
+    text-decoration: none;
+  }
+}
+
+.iasWidget-img {
+  display: block;
+  height: auto;
+  margin: 0 auto;
+  width: 25%;
+}
+
+.iasWidget-row {
+  @include make-row;
+}
+
+@media only screen and (min-width: $screen-sm-min) {
+  .iasWidget-img {
+    width: 33.3%;
+  }
+}
--- a/styleguide/_themes/derek/scss/patterns/grid/ias.scss
+++ b/styleguide/_themes/derek/scss/patterns/grid/ias.scss
@@ -40,11 +40,11 @@
   @include flex-box;
 }
 
-.ias5-col-push {
+.ias-col5-push {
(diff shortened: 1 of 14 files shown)


styleguide/_themes/derek/scss/patterns/grid/ias.scss
```diff
@@ -0,0 +1,104 @@
+.ias-container {
+  @include container-fixed;
```
WidgetsBurritos: If you're going to use the container-fixed mixin, you'll need to set everything for the various viewports to emulate bootstrap behavior.  This will eventually get combined into a single mixin, but for now add that. You can see how john did it here: 

https://github.com/rackerlabs/zoolander/pull/173/files#diff-df6cfdc6c265be1c82cdd82efd2ac4d8R18

styleguide/derek/_partials/solutions/services-5-col.ejs
```diff
@@ -1,27 +1,24 @@
+
```
WidgetsBurritos: I'd say no blank lines to start files.

styleguide/derek/_partials/solutions/services-6-col.ejs
```diff
@@ -1,32 +1,27 @@
...
-<div class="container">
-  <div class="row">
-    <div class="col-md-12">
-      <h2 class="center">6 columns</h2>
-    </div>
-  </div>
-  <div class="row row-eq-height">
-    <div class="col-sm-4 clickarea">
+<div class="ias-container">
+      <h2 class="ias-title">6 columns</h2>
```
WidgetsBurritos: Fix indention here.

styleguide/_themes/derek/scss/patterns/grid/ias.scss
```diff
@@ -0,0 +1,112 @@
...
+  max-width: 960px;
+  padding-bottom: 60px;
```
WidgetsBurritos: Can you add a `$standard-padding` variable in `_themes/global/scss/modifiers.scss` with a value of `40px 15px 60px;` and then reference it here instead of using top/bottom padding, so there's no ambiguity as to what the padding is here?

styleguide/derek/_partials/grid/widget/ias.ejs
```diff
@@ -2,24 +2,13 @@
...
-    <div class="col-md-12">
-      <h3 class="center">Name with only 30 characters</h3>
-    </div>
-    <div class="col-xs-3 center-block">
-      <hr>
-    </div>
+<a class="iasWidget iasWidget-hover" href="http://www.google.com">
+  <img class="iasWidget-img" src="https://placeholdit.imgix.net/~text?txtsize=30&txt=%5Bplaceholder%5D&w=200&h=150">
+  <div class="iasWidget-row">
     <div class="col-md-12">
```
WidgetsBurritos: Change this to an `iasWidget-col-full` class and then add mixin to emulate behavior.

styleguide/derek/_partials/grid/widget/ias.ejs
```diff
@@ -2,24 +2,13 @@
...
-    </div>
-    <div class="col-xs-3 center-block">
-      <hr>
-    </div>
-    <div class="col-md-12">
-      <p class="center"><%- desc %></p>
-    </div>
-    <div class="bottom">
-      <p class="center red semi-bold">More Info</p>
+<a class="iasWidget iasWidget-hover" href="http://www.google.com">
```
WidgetsBurritos: Can you change this href to `javascript:void(0)`

WidgetsBurritos:
👍

styleguide/_themes/derek/scss/patterns/grid/ias.scss
```diff
@@ -0,0 +1,115 @@
...
+  @include flex-box;
+}
+
+.ias-col-third {
+  @include make-xs-column(12);
+  @include make-sm-column(4);
+  @include flex-box;
+}
+
+.ias5-col-push {
```
amberRrucker: ias5? if this is the same "module" as ias then doesnt it go against maintainable to have the prepended class different?

styleguide/_themes/derek/scss/patterns/grid/ias.scss
```diff
@@ -0,0 +1,115 @@
...
+  @include flex-box;
+}
+
+.ias-title {
+  text-align: center;
+}
+
+.iasWidget {
+  padding: 20px 10px 0;
+  width: 100%;
```
amberRrucker: you can use width: $full-width if you would like here.

styleguide/_themes/derek/scss/patterns/grid/ias.scss
```diff
@@ -0,0 +1,115 @@
...
+}
+
+.iasWidget-body {
+  color: $darkest-gray;
+  font-family: $copy;
```
amberRrucker: isnt this all standard? all our standard copy is copy darkest gray, 300 and what not. if we change that globally we're going to want this to change as well.

styleguide/_themes/derek/scss/patterns/grid/ias.scss
```diff
@@ -0,0 +1,115 @@
...
+  text-align: center;
+}
+
+.iasWidget-col-full {
+  @include make-md-column(12);
+}
+
+.iasWidget-cta {
+  color: $rackspace-red;
+  font-family: $copy;
```
amberRrucker: does this really need to be defined?

styleguide/_themes/derek/scss/patterns/grid/ias.scss
```diff
@@ -0,0 +1,115 @@
...
+}
+
+.iasWidget-header {
+  color: $darkest-gray;
+  font-family: $heading;
+  font-size: 1.25em;
+  font-weight: 400;
+  line-height: 1.25em;
+  margin: 10px 0;
+  text-align: center;
```
amberRrucker: the only thing that should be non default for the h3 is the centering. We shouldnt define all of the rest because when the h3 changes, we're going to want this to change as well. If we switch to a new font (a reality very soon) then we'll need the font-size, weight and line height to change as well. If we define this so heavily it will complicate that process quite a bit.

Also, the font weight is incorrect. This needs to be 300 or removed. (I recommend removed)

amberRrucker:
I feel like previously the learn more link appeared at the bottom of the equal height container. I would really like to return to that, visually.

amberRrucker:
+1 looks good

WidgetsBurritos:
Code still looks good after changes.   +1

PR #179, opened 2016-07-12
rsweb-6762: fixing subnav active state

SamuelRandall:
https://jira.rax.io/browse/RSWEB-6762


bighappyface:
Please exclude the changes in dist

PR #181, opened 2016-07-13
table-styling

amberRrucker:
https://jira.rax.io/browse/RSWEB-6675
Added some table styling and put in some examples for that. Didn't add to the examples page because of the front end changes going through review anyways. Don't want to add it twice.

@bighappyface, @WidgetsBurritos I'm curious about if you're okay with my treatment on horizontal-scroll. Im pretty sure in maintainable they would say each element should have an additional class put on it to handle horizontal scrolling changes but I would really want to avoid doing that. Especially since tables wont be a pattern and will be utilized by ECMs on occasion.

That's a group decision though.
I also took the opportunity to switch up the hex values in forms.scss to get us down to 120 sass-lint errors.

### DIFF
--- a/styleguide/_themes/global/scss/forms.scss
+++ b/styleguide/_themes/global/scss/forms.scss
@@ -28,7 +28,7 @@ table {
   &.checkbox,
   &.radio-inline,
   &.checkbox-inline {
-    color: #eb0022;
+    color: $brand-danger;
   }
 
   .checkbox,
@@ -37,12 +37,12 @@ table {
   .help-block,
   .radio,
   .radio-inline {
-    color: #eb0022;
+    color: $brand-danger;
   }
 
   .form-control {
     background-color: rgba(235, 0, 34, .1);
-    border-color: #eb0022;
+    border-color: $brand-danger;
   }
 }
 
@@ -51,7 +51,7 @@ table {
   &.checkbox-inline,
   &.radio,
   &.radio-inline {
-    color: #008500;
+    color: $brand-success;
   }
 
   .checkbox,
@@ -60,12 +60,12 @@ table {
   .help-block,
   .radio,
   .radio-inline {
-    color: #008500;
+    color: $brand-success;
   }
 
   .form-control {
     background-color: rgba(0, 133, 0, .1);
-    border-color: #008500;
+    border-color: $brand-success;
   }
 }
 
@@ -75,7 +75,7 @@ table {
   &.radio,
   &.radio-inline {
     label {
-      color: #eb6800;
+      color: $brand-warning;
     }
   }
 
@@ -85,12 +85,12 @@ table {
   .help-block,
   .radio,
   .radio-inline {
-    color: #eb6800;
+    color: $brand-warning;
   }
 
   .form-control {
     background-color: rgba(235, 104, 0, .1);
-    border-color: #eb6800;
+    border-color: $brand-warning;
   }
 }
 
--- a/styleguide/_themes/global/scss/tables.scss
+++ b/styleguide/_themes/global/scss/tables.scss
@@ -1,0 +1,105 @@
+$table-padding: 20px 5px;
+
+.productTable {
+  border-collapse: collapse;
+  width: 100%;
+}
+
+.productTable-feature,
+.productTable-featureItem,
+.productTable-name {
+  border: 1px solid $light-gray;
+}
+
+.productTable-name {
+  @include heading;
+  background-color: $darkteal;
+  color: $white;
+  padding: $table-padding;
+  text-align: center;
+}
+
+.productTable-feature {
+  @include subhead;
+  color: $darkteal;
+  padding: $table-padding;
+  text-align: center;
+}
+
+.productTable-featureItem {
+  &::before {
+    @include center-block;
+    font-family: FontAwesome;
+    padding: 10px 0;
+    text-align: center;
+  }
+}
+
+.offered {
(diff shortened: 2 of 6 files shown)


styleguide/_themes/global/scss/tables.scss
```diff
@@ -0,0 +1,105 @@
...
+    justify-content: center;
+
+    :first-child {
+      background: $light-gray;
+      display: block;
+      width: 100%;
+    }
+  }
+
+  .horizontal-scroll {
```
WidgetsBurritos: I haven't thoroughly thought this through, but maybe we could write `horizontal-scroll($prefix)` as a mixin instead and then just have the `.horizontal-scroll` include that mixin.  Then we could have supply the prefix for for that table for maintainable-specific examples, while still having an option for ECMs.

styleguide/_themes/global/scss/forms.scss
```diff
@@ -37,12 +37,12 @@ table {
   .help-block,
   .radio,
   .radio-inline {
-    color: #eb0022;
+    color: $brand-danger;
   }
 
   .form-control {
     background-color: rgba(235, 0, 34, .1);
```
WidgetsBurritos: While you're at it, can we switch these rgba() values to variable names too, to clean up these sass lint errors? 

(We may need to figure out a naming convention for alpha-based colors as well)

styleguide/_themes/global/scss/tables.scss
```diff
@@ -0,0 +1,105 @@
...
+    display: flex;
+    flex-flow: row wrap;
+    justify-content: space-around;
+  }
+
+  .productTable-name {
+    display: flex;
+    flex-flow: row;
+    flex-grow: 1;
+    font-size: .70em;
```
WidgetsBurritos: From sass-lint:  `70:16  warning  Don't include trailing zeros on numbers  no-trailing-zero`

styleguide/_themes/global/scss/tables.scss
```diff
@@ -0,0 +1,105 @@
...
+      background: $light-gray;
+      display: block;
+      width: 100%;
+    }
+  }
+
+  .horizontal-scroll {
+    display: table;
+
+    .productTable-row{
```
WidgetsBurritos: From sass-lint:
`95:21  warning  Whitespace required before {             space-before-brace`

WidgetsBurritos:
Just a general commentary on sass lint. While lowering the sass-lint errors, in this PR, you did introduce a few new ones in there, that didn't prevent you from committing, because you were below the threshhold.  It's important to check the sass linter for errors on changes you made, because our goal is to eventually get that number down to zero.

WidgetsBurritos:
Maybe while you're at it, once you cleanup the new sass lint issues here, can you modify line 19 in `package.json` to whatever the new limit is based on your PR: 
```"sass-lint": "sass-lint -v --max-warnings 127",```

styleguide/_themes/global/scss/forms.scss
amberRrucker: @WidgetsBurritos do we want to have a group chat about this or go ahead and change it? its $brand-danger with an opacity on it. not sure if we wanna set a $brand-danger-bg or something.

styleguide/_themes/global/scss/tables.scss
amberRrucker: @WidgetsBurritos I'm gonna wait for additional recommendations or a final call or maybe just another plus one to do this one.

styleguide/_themes/global/scss/forms.scss
WidgetsBurritos: Yeah probably would be a good idea to at least get @bighappyface and @SyntaxStacks's feedback on this.

styleguide/_themes/global/scss/tables.scss
WidgetsBurritos: @amberRrucker Definitely makes sense. Like I said, I haven't fully processed the concept, but I think it at least gives us an option, but I'd be interested in what the others have to say about this.

styleguide/_themes/global/scss/forms.scss
SyntaxStacks: I appreciate the eye for detail, but if the errors are already accounted for we can come back to fix them up in a PR specifically for that. We added the warnings threshold to avoid hold ups like this in code review.

WidgetsBurritos: Fair enough.

SyntaxStacks:
Added table styling to the examples page

SyntaxStacks:
screen shot 2016-07-14 at 8 39 12 am
screen shot 2016-07-14 at 8 39 33 am
screen shot 2016-07-14 at 8 40 11 am

SyntaxStacks:
things get a bit tight in tablet on longer tables

amberRrucker:
@SyntaxStacks I adjusted the padding to account for longer tables in tablet. It would make me berry berry happy if we could get this one out so if we could get some review and +1s that would be super lovely. @WidgetsBurritos

WidgetsBurritos:
@amberRrucker Can you add a link to the tables page in the examples page?  Right now there doesn't appear to be a way to get there, unless you manually type in `tables.html`.

Unless your reorg PR will address this issue differently, in which case I'd say you're fine without it.

SyntaxStacks:
The reorg PR will address the tables pattern example :+1:

PR #182, opened 2016-07-13
doing the far from overdo homepage sass refactor

amberRrucker:
I don't have a PR for this one. Sorry everyone. But it had to be done. It was really garbage-y and hanging around.


styleguide/derek/examples/homepage.ejs
```diff
@@ -2,182 +2,190 @@
...
-    <div class="container standard-padding-full">
-      <div class="col-md-8">
-        <h2>Expertise comes standard.<br/> Choose your Managed Cloud.</h2>
-        <p class="left white">Trusted by more than 60% of the Fortune 100. Serving 300,000+ businesses globally. A leader in the Gartner Magic Quadrant for cloud-enabled managed hosting.</p>
-        <div class="row">
-          <div class="col-lg-4 col-md-6">
-            <a href="#chat" class="button lead">Speak with a specialist</a>
-          </div>
+<div class="container-fluid homepage-hero homepageHero">
+  <div class="container standard-padding-full">
```
WidgetsBurritos: Can you fix these nested containers?

styleguide/derek/examples/homepage.ejs
```diff
@@ -2,182 +2,190 @@
...
 
-  <!-- Hosting Options -->
-  <div class="container standard-padding" id="hosting-options">
-    <div class="row">
-      <div class="col-md-12 lead">
-        <h3>We're the largest managed cloud provider, offering you expertise across the world's leading clouds</h3>
-      </div>
+<!-- Hosting Options -->
+<div class="container standard-padding hostingOptions" id="hosting-options">
+	<div class="row half-padding-full">
```
WidgetsBurritos: These tab sizes are larger than rest of file.  Can we use 2 spaces across the board?

amberRrucker: Sure yeah. I pulled them in from the Drupal node. I didn't check to see if they were different. Whoops.

styleguide/derek/examples/homepage.ejs
```diff
@@ -2,182 +2,190 @@
...
-            </div>
-            <div class="col-xs-3 center-block">
-              <hr>
-            </div>
-            <div class="col-md-12">
-              <p class="center">Co-founder with NASA; the most experience in the world by 100-times</p>
+  </div>
+  <div class="row half-padding-full">
+    <div class="col-md-11 center-block">
+    <a href="[rs_link:managed-security]" class="solution solutions-hoverArea">
```
WidgetsBurritos: Probably wanna pull the Drupal token out of here

amberRrucker: Do we necessarily need to? It would be easier to updatecode for Drupal if we leave it in and it isn't a reusable pattern. Just a development environment/example right? Shouldn't we try to use the tokens in that case? So we don't dev and accidentally leave that out in future iterations?

WidgetsBurritos: Well I guess we can, but I thought the point of zoolander was to de-drupalize things.   Also its kind of weird when I click on a link in zoolander and get this: 

image

I kinda like just using `javascript:void(0)` for all links, unless they actually go somewhere on zoolander.  But that may be just me.

PR #188, opened 2016-07-19
feat(solutions):blog-feed

amberRrucker:
https://jira.rax.io/browse/RSWEB-6682

Adding dynamic blog feed front end code and documentation.


styleguide/_themes/derek/scss/modifiers.scss
```diff
@@ -45,6 +45,13 @@ $body-bg:  #fff !default;
...
 
 //centered-divs
+@mixin center-block {
+  display: block;
+  float: none;
+  margin: 0 auto;
+}
+
+//making sure to support the class
 .center-block {
```
WidgetsBurritos: If you're gonna do this, should `.center-block` then include this mixin instead?

amberRrucker: Good call. Fixed it!

styleguide/_themes/derek/scss/patterns/blog-feed.scss
```diff
@@ -0,0 +1,71 @@
+@mixin blogFeed-textElements {
```
WidgetsBurritos: Is this mixin going to be reused in the future?  If it's only referenced in one place, I don't necessarily see the need for a mixin, but if you have future plans for it, then that's okay with me.  Just checking.

amberRrucker: Well its used in almost all of these elements. I used it so that in the future if we want to change how this text appears we can just change it here instead of going through all of those. I wouldve just used a class and extended it but sass-lint says thats a no-go. I included it within this file instead of a global file to avoid anyone using it for something other than this. so is there a better way? I felt like this was in the spirit of maintainable. Easy to update, module specific. ya know that stuff.

styleguide/_themes/global/scss/vendor_prefixes.scss
```diff
@@ -41,3 +41,7 @@
   display: -webkit-flex;
   display: flex;
 }
+
+@mixin sepia-filter {
+  -webkit-filter: sepia(.9);
```
WidgetsBurritos: Shoudl this also have a `filter: sepia(.9);`?

styleguide/derek/_partials/grid/widget/blog-feed-item.ejs
```diff
@@ -0,0 +1,23 @@
+<!-- Blog Feed Pattern -->
+
+<div class="blogFeed-item">
+  <div class="row">
```
WidgetsBurritos: Probably should use `blogFeed-row` instead of default bootstrap rows.

amberRrucker: Do we need to do that if there is no special css, js or classing there at all? Wouldnt it just be adding a class for no reason?

styleguide/_themes/derek/scss/patterns/blog-feed.scss
```diff
@@ -0,0 +1,71 @@
...
+
+.blogFeed-title {
+  color: $medium-dark-gray;
+}
+
+.blogFeed-author {
+  font-size: .7em;
+
+  &::before {
+    content: 'By: ';
```
WidgetsBurritos: Content inside css won't work with localization .

styleguide/derek/_partials/solutions/blog-feed.ejs
```diff
@@ -0,0 +1,28 @@
+<!--  IMPLEMENTATION GUIDELINES -->
+
+<!--
+Limit to excerpt 150 characters - I cant implement with pure css because it needs to be multiple lines. Should culminate in an ellipsis.
```
WidgetsBurritos: That's fine .

.sass-lint.yml
```diff
@@ -13,3 +13,7 @@ rules:
     - 1
     - convention: ^([a-z0-9]+([A-Z][a-z0-9]+)*)(-{1}([a-z0-9]+([A-Z][a-z0-9]+)*)){0,2}$
     - convention-explanation: 'Classes should be camelCase and separated by no more than 2 hyphens'
+  mixin-name-format:
+    - 1
+    - convention: ^([a-z0-9]+([A-Z][a-z0-9]+)*)(-{1}([a-z0-9]+([A-Z][a-z0-9]+)*)){0,5}$
+    - convention-explanation: 'Classes should be camelCase and separated by no more than 2 hyphens'
```
WidgetsBurritos: This comment should be changed to 5 hyphens.  Sorry I missed that earlier.

styleguide/_themes/derek/scss/patterns/blog-feed.scss
WidgetsBurritos: This is fine.

styleguide/_themes/derek/scss/patterns/blog-feed.scss
WidgetsBurritos: For clarification: any text content inside css will not go through smartling and thus not get translated.  That text should be hardcoded into HTML, not placed inside css.

styleguide/derek/_partials/grid/widget/blog-feed-item.ejs
WidgetsBurritos: This helps prevent other conflicts with something else that may come along the way.  `.row` is such a generic class name, we're just asking for some crappy module plugin to break things.

styleguide/_themes/derek/scss/patterns/blog-feed.scss
amberRrucker: @ejchiefs are we supporting translation for the blog content block?

styleguide/_themes/derek/scss/patterns/blog-feed.scss
amberRrucker: Your response is anti-climactic.

styleguide/_themes/derek/scss/patterns/blog-feed.scss
bighappyface: Maybe not the very first use of this but l10n should always be up front. We can just pull it from css and prepend it in Drupal

styleguide/_themes/derek/scss/patterns/blog-feed.scss
WidgetsBurritos: You're welcome.

bighappyface:
+1

PR #195, opened 2016-07-21
RSWEB-6880: set width for feature

writes:
https://jira.rax.io/browse/RSWEB-6880

updates table feature to break text


WidgetsBurritos:
:+1:

amberRrucker:
+1 perfect. We need to also set word wrapping on long titles for mobile but we can tackle that in another PR. this does the job for digital for sure.

PR #200, opened 2016-07-28
zoolander(bugfix): adjusted ticker string so that it won't jitter around

runrog:
https://jira.rax.io/browse/RSWEB-6926

### DIFF
--- a/styleguide/_themes/derek/scss/components/ticker.scss
+++ b/styleguide/_themes/derek/scss/components/ticker.scss
@@ -55,9 +55,10 @@ $ticker-span-bg: rgba(200, 0, 0, .5);
   font-size: 29px;
   font-weight: 700;
   letter-spacing: 6px;
-  margin: 30px;
+  margin: 30px auto;
   padding: 15px;
   text-align: left;
+  width: 265px;
 }
 
 .ticker-image {


WidgetsBurritos:
+1 looks okay for me.

styleguide/_themes/derek/scss/components/ticker.scss
```diff
@@ -55,9 +55,10 @@ $ticker-span-bg: rgba(200, 0, 0, .5);
   font-size: 29px;
   font-weight: 700;
   letter-spacing: 6px;
-  margin: 30px;
+  margin: 30px auto;
   padding: 15px;
   text-align: left;
+  width: 265px;
```
amberRrucker: i still dont love the idea of used a fixed pixel width and would prefer to use the make-(size)-col to vary the widths and break points. @bighappyface and @SyntaxStacks thoughts?

SyntaxStacks: I didn't see this! Yeah i'd rather have the columns in place.

amberRrucker: REVERT REVERT! haha. I thought we were doing 2 plus 1s for every release now.

PR #199, opened 2016-07-28
RSWEB-6688: Add Events into Zoolander

WidgetsBurritos:
Jira: https://jira.rax.io/browse/RSWEB-6688

This PR modifies the events to make them bootstrappy and maintainable (and also look better).

PR #202, opened 2016-07-29
RSWEB-6688: Map canvas does need predefined dimensions after all

WidgetsBurritos:
Jira: https://jira.rax.io/browse/RSWEB-6688

This PR is a follow up to #199.  As I did some initial testing in www, based on the way the JS was setup, it turns out we do need a predefined canvas size for the map after all.

### DIFF
--- a/styleguide/_themes/derek/scss/snowflakes/events.scss
+++ b/styleguide/_themes/derek/scss/snowflakes/events.scss
@@ -276,6 +276,11 @@
   margin-bottom: 2rem;
 }
 
+.event-mapCanvas {
+  height: 250px;
+  width: 100%;
+}
+
 .event-row {
   @include make-row;
 }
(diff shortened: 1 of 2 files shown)


amberRrucker:
booooooo www but understandable +1

writes:
+1 looks good to me

SyntaxStacks:
screen shot 2016-07-29 at 10 29 11 am
Looks interesting in tablet

writes:
Will these actually be images or will it be using the Google map api? I assumed the api would handle sizing issues but maybe not?

SyntaxStacks:
If we're using the google maps API that scaling should be a non issue

WidgetsBurritos:
@SyntaxStacks Correct. The image gets replaced by an actual JS map so the scale looks okay.  I just use the image here as a placeholder.

PR #203, opened 2016-07-29
feat(landing-pages):first pass at landing page options

amberRrucker:
as well as video embed, form for lp, and documentation around those pages.
https://jira.rax.io/browse/RSWEB-6795


styleguide/_themes/derek/scss/patterns/videoEmbed.scss
```diff
@@ -0,0 +1,14 @@
...
+.videoEmbed-row {
+  @include make-row;
+}
+
+// when the make column is used in the sass then it overrides any center-block declaration with a float:left ??
+.videoEmbed-column {
+  @include make-md-column(6)
+  @include center-block;
+  padding: $half-padding-full;
+}
```
amberRrucker: @SyntaxStacks and @WidgetsBurritos  do you have any idea how we can have both?

WidgetsBurritos: Do you need a semicolon at the end of `@include make-md-column(6)`?

Not sure if that affects anything, but worth a shot.

WidgetsBurritos: Can you just do an `@include make-md-column-push(3);` on here as well?

styleguide/_themes/derek/scss/_main.scss
```diff
@@ -10,6 +10,7 @@
 @import 'components/back-to-top';
 @import 'components/regionSelector';
 @import 'components/ticker';
+@import 'patterns/videoEmbed';
```
WidgetsBurritos: This should be in alphabetical order?

styleguide/_themes/global/scss/modifiers.scss
```diff
@@ -71,6 +71,11 @@ img { max-width: 100%; }
   margin: 0 auto;
 }
 
+@mixin center-block {
```
WidgetsBurritos: So now that you're no longer using this mixin, we probably don't need to include it in this PR.  If we want to create this mixin in the future when we actually use it, that seems okay, but for now, I'd say scrap it.

WidgetsBurritos:
👍

ejchiefs:
👍

PR #204, opened 2016-07-29
style(zoolander):h1-swap

amberRrucker:
https://jira.rax.io/browse/RSWEB-6940
Switches h1 to lead position and h2 to banner position. Need to switch h2 class for our drupal pattern banners to banner-headline, to avoid the h2 resizing issue, since h2 is our designated river title size.

View title in banner cases here:
http://localhost:9000/derek/assets/photography

typography cases here: 
http://localhost:9000/derek/assets/typography

and the lead/heading changes here:
http://localhost:9000/derek/examples/solutions

PR #206, opened 2016-08-01
RSWEB-6688: Add banner text.

WidgetsBurritos:
Per my meeting with the events team last week, I needed to add text to the banner region on the events page. I've gone ahead and added it here and converted it to maintainable css.


ejchiefs:
👍

styleguide/_themes/global/scss/banners.scss
```diff
@@ -1,22 +1,31 @@
...
   background-size: cover;
   color: $white;
   font-family: $heading;
+}
+
+$banner-height-large: 375px;
+$banner-height-medium: 225px;
+$banner-text-margin-large: 18%;
+$banner-text-margin-medium: 10%;
+
```
amberRrucker: this stuff is going to change for the imac banners but we will still have to support the legacy stuff. its going to get kind of interesting but this will eventually be suuuuuper useful i think.

amberRrucker:
This stuff will need a www implementation task to get it through, it's also an overlap/double work on the banner things.  This may just complicate the release of the h1/h2 switch when we can just stick to the original scope of adding in banner text and make these changes in the imac banner task. If we do this stuff to the banner structure and classing everything we just reviewed for the h1/h2 switch is going to need to be redone. 
@ejchiefs @WidgetsBurritos

WidgetsBurritos:
@amberRrucker EJ and I discussed that earlier. 

The existing banner text behavior doesn't work here, because we need to address SEO concerns as well. Contextually, this text needs to be an `<h1>` on the events page and a non-h1 on the individual event pages. That's why it is necessary to hard code font size in the `.events-bannerTitle` class (as its not pulling in side from the H1). 

That said, the creation of the mixins/variables here should help reduce the amount of double work that happens as a result of this. Also, based on our previous discussions last week about using bootstrap's `$font-size-h1` variable in the future, it should be easy enough to set these font sizes in the future as well. 

To solve this problem we're gonna need custom CSS on the text one way or another, so I'm not sure that's a viable strategy in the interim.

WidgetsBurritos:
Adding the hold tag here, because we need to get the H1-swap PR (#204) out first.

writes:
👍

PR #210, opened 2016-08-09
feat(zoolander):imac-banners

amberRrucker:
https://jira.rax.io/browse/RSWEB-6941

### DIFF
--- a/styleguide/_layouts/default.jade
+++ b/styleguide/_layouts/default.jade
@@ -20,6 +20,7 @@ html(lang="en")
     link(href="https://maxcdn.bootstrapcdn.com/font-awesome/4.6.1/css/font-awesome.min.css" rel="stylesheet")
     script(src="https://code.jquery.com/jquery-1.12.3.min.js" integrity="sha256-aaODHAgvwQW1bFOGXMeX+pC4PZIPsvn2h1sArYOhgXQ=" crossorigin="anonymous")
     script(src="https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/zoolander/modernizr.min.js")
+    script(src="https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/zoolander/jquery.fittext.js")
 
   .container-fluid
     .row.row-eq-height
@@ -71,3 +72,5 @@ html(lang="en")
   script(src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/js/bootstrap.min.js" integrity="sha384-0mSbJDEHialfmuBBQP6A4Qrprq5OVfW37PRR3j5ELqxss1yVqOtnepnHVP9aJ7xS" crossorigin="anonymous")
   script(src="#{base}js/zoolander.js")
   script(src="https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/zoolander/bootstrap-tabcollapse.js")
+  script.
+    $(".imacBanner-headline").fitText(1.1, { maxFontSize: '45px' });
--- a/styleguide/_themes/derek/scss/snowflakes/homepage.scss
+++ b/styleguide/_themes/derek/scss/snowflakes/homepage.scss
@@ -18,7 +18,7 @@
     @extend .white;
     @extend .left;
   }
-  p{
+  p {
     letter-spacing: 1px;
     line-height: 24px;
     font-size: 16px;
@@ -26,7 +26,7 @@
   }
 }
 
-#hosting-options{
+#hosting-options {
   text-align:center;
   .product{
     p{ margin-bottom:30px; }
--- a/styleguide/_themes/global/scss/banners.scss
+++ b/styleguide/_themes/global/scss/banners.scss
@@ -36,31 +36,127 @@ $banner-text-margin-medium: 10%;
 
 .banner-headline {
   @include heading($white);
-  font-size: 3em;
-  line-height: 1em;
+  font-size: 4rem;
+  line-height: 3.75rem;
   text-align: center;
 }
 
 
-.imacBanner-container {
-  background-color: $brand-primary;
+.imacBanner {
+  background-image: url('https://752f77aa107738c25d93-f083e9a6295a3f0714fa019ffdca65c3.ssl.cf1.rackcdn.com/bananers/backgrounds/o365-overview.jpg');
+  background-size: cover;
   color: $white;
+  margin-top: 50px;
+  overflow: hidden;
+  padding: 50px 0;
+}
 
-  &.large {
-    height: 300px;
-    padding-top: 30px;
-  }
+.imacBanner-container {
+  @include make-container;
+  @include fadein(1s);
+}
+
+.imacBanner-textRow {
+  @include make-row;
+}
+
+.imacBanner-headlineLarge {
+  @include make-sm-column(7);
+}
+
+.imacBanner-headlineMedium {
+  @include make-sm-column(6);
+}
+
+.imacBanner-headlineSmall {
+  @include make-sm-column(5);
 }
 
 .imacBanner-headline {
   @include heading($white);
+  // will be removing this line when imac comes out because we're actually going to be switching the heading mixin to use Khand anyways
+  font-family: 'Khand';
   font-size: 4rem;
   font-weight: 700;
-  line-height: 3.75rem;
+  line-height: .9em;
   margin: 0;
 }
 
-.imacBanner-text {
+.imacBanner-textContainer {
   border: 3px solid $white;
+  margin: 0 auto;
   padding: 15px;
+  position: relative;
+}
+
+.imacBanner-buttonRow {
+  @include make-row;
+  padding-top: 20px;
+}
+
+.imacBanner-button {
+  @include make-sm-column(3);
+  @include button;
+  background-color: $brand-primary;
+  color: $white;
+  font-weight: 600;
+  margin-left: 20px;
+
+  &:hover {
+    background-color: darken($brand-primary, 5%);
+    color: $white;
+  }
+}
+
+.imacBanner-buttonIcon {
+  @include text-stroke(1px, $white);
+  color: $white;
+  float: right;
+  margin-top: 3px;
+}
+
+.line {
+  @include rotate(45deg);
+  background-color: $white;
+  display: block;
+  height: 45rem;
+  position: absolute;
+  width: 3px;
+
+  &.left-top {
+    left: -3px;
+    top: -2px;
+    transform-origin: 1px 0;
+  }
+
+  &.left-bottom {
+    left: -2px;
+    top: 100%;
+    transform-origin: 1px 0;
+  }
+
+  &.right-bottom {
+    left: 100%;
+    top: 100%;
+    transform-origin: -1px 2px;
+  }
+}
+
+@media only screen and (max-width: $screen-sm-max) {
+  .imacBanner {
+    margin-top: 20px;
+  }
+}
+
+@media only screen and (max-width: $screen-xs-max) {
+  .imacBanner {
+    background-color: $black;
+    background-image: none;
+    padding: 40px 15px;
+  }
+
+  .imacBanner-button {
+    margin-left: 20px;
+    width: 70%;
+  }
 }
--- a/styleguide/_themes/global/scss/buttons.scss
+++ b/styleguide/_themes/global/scss/buttons.scss
@@ -6,10 +6,10 @@
   display: inline-block;
   font-size: .85em;
   font-weight: 600;
-  padding: 15px 20px;
+  padding: 8px 10px;
   text-align: center;
   text-decoration: none;
-  text-transform: uppercase;;
+  text-transform: uppercase;
   width: 100%;
 
   &:hover {
--- a/styleguide/_themes/global/scss/vendor_prefixes.scss
+++ b/styleguide/_themes/global/scss/vendor_prefixes.scss
@@ -41,6 +41,12 @@
   border-radius: $radius;
 }
 
+@mixin rotate($degrees) {
+  -webkit-transform: rotate($degrees);
+  -ms-transform: rotate($degrees);
+  transform: rotate($degrees);
+}
+
 @mixin flex-box {
   display: -webkit-box;
   display: -ms-flexbox;
@@ -51,4 +57,45 @@
 @mixin sepia-filter {
   -webkit-filter: sepia(.9);
   filter: sepia(90%);
+}
+
+@mixin text-stroke($width, $color) {
+  -webkit-text-stroke: $width $color;
+}
+
+@mixin fadein($duration) {
+  -webkit-animation: fadein $duration;
+  -moz-animation: fadein $duration;
+  -ms-animation: fadein $duration;
+  -o-animation: fadein $duration;
+  animation: fadein $duration;
+}
+
+@keyframes fadein {
+    from { opacity: 0; }
+    to   { opacity: 1; }
+}
+
+/* Firefox < 16 */
+@-moz-keyframes fadein {
+    from { opacity: 0; }
+    to   { opacity: 1; }
+}
+
+/* Safari, Chrome and Opera > 12.1 */
+@-webkit-keyframes fadein {
+    from { opacity: 0; }
+    to   { opacity: 1; }
+}
+
+/* Internet Explorer */
+@-ms-keyframes fadein {
+    from { opacity: 0; }
+    to   { opacity: 1; }
+}
+
+/* Opera < 12.1 */
+@-o-keyframes fadein {
+    from { opacity: 0; }
+    to   { opacity: 1; }
 }
--- a/styleguide/derek/_partials/solutions/header-mediumIMAC.ejs
+++ b/styleguide/derek/_partials/solutions/header-mediumIMAC.ejs
@@ -1,9 +1,9 @@
-<!--Header Long-->
+<!--Header Medium-->
 <div class="imacBanner">
   <div class="imacBanner-container">
     <div class="imacBanner-textRow">
-      <div class="imacBanner-textContainer imacBanner-headlineLarge">
-        <h2 class="imacBanner-headline">we guarantee a 99.99% openstack API uptime.<br/>And yes, that .01% does keep us up at night</h2>
+      <div class="imacBanner-textContainer imacBanner-headlineMedium">
+        <h2 class="imacBanner-headline">the cloud is a mystery wrapped in a riddle.<br/>We can unwrap it for you</h2>
         <span class="line left-top"></span>
         <span class="line left-bottom"></span>
         <span class="line right-bottom"></span>
--- a/styleguide/derek/_partials/solutions/header-smallIMAC.ejs
+++ b/styleguide/derek/_partials/solutions/header-smallIMAC.ejs
@@ -1,0 +1,13 @@
+<!--Header Short-->
+<div class="imacBanner">
+  <div class="imacBanner-container">
+    <div class="imacBanner-textRow">
+      <div class="imacBanner-textContainer imacBanner-headlineSmall">
+        <h2 class="imacBanner-headline">The cloud runs on its own,<br/>with lots of help</h2>
+        <span class="line left-top"></span>
+        <span class="line left-bottom"></span>
+        <span class="line right-bottom"></span>
+      </div>
+    </div>
+  </div>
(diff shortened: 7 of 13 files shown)


styleguide/_themes/global/scss/vendor_prefixes.scss
```diff
@@ -41,6 +41,12 @@
   border-radius: $radius;
 }
 
+@mixin transform($degrees) {
```
SamuelRandall: Maybe call this `rotate` instead of `transform` in case we ever want to make a mixin for other transform properties?

styleguide/derek/examples/imac.jade
```diff
@@ -0,0 +1,202 @@
...
+          h2.imacBanner-headline.imacBanner-headlineMedium spoiler alert: <br/>The cloud isn’t all sunshine and rainbows
+          .line.left-top
+          .line.left-bottom
+          .line.right-bottom
+      .imacBanner-buttonRow
+        a.button.imacBanner-button Read Our Case Studies
+          i.fa.fa-angle-right.imacBanner-buttonIcon
+
+script.
+  $(".imacBanner-headlineLarge").fitText(1.1, { maxFontSize: '45px' });
```
SamuelRandall: If this script is being called on each individual '.ejs' file, does it need to be called here as well?

styleguide/_themes/global/scss/banners.scss
```diff
@@ -36,31 +36,125 @@ $banner-text-margin-medium: 10%;
...
 
+.imacBanner-container {
+  @include make-container;
+}
+
+.imacBanner-textRow {
+  @include make-row;
+}
+
 .imacBanner-headline {
```
SamuelRandall: We could combine `imacBanner-headlineLarge` and `imacBanner-headlineMedium` into here so that we don't have multiple classes on those elements. That way we avoid using a class that has no styling attached to it and is only used for jQuery referencing.

styleguide/derek/_partials/solutions/header-largeIMAC.ejs
```diff
@@ -1,18 +1,20 @@
...
-        <a class="button banner">Choose A Plan</a>
-      </div>
+    <div class="imacBanner-buttonRow">
+      <a class="button imacBanner-button">Read Our Case Studies  <i class="fa fa-angle-right imacBanner-buttonIcon"></i></a>
     </div>
   </div>
 </div>
+
+<script type="text/javascript">
+	$(".imacBanner-headlineLarge").fitText(1.1, { maxFontSize: '45px' });
```
SamuelRandall: Also, it seems that every one of these is the same, could we not just apply this script to the `imacBanner-headline` class instead of using these non-styled class names?

styleguide/_themes/global/scss/banners.scss
```diff
@@ -36,31 +36,125 @@ $banner-text-margin-medium: 10%;
...
   @include heading($white);
+  font-family: 'Khand';
   font-size: 4rem;
   font-weight: 700;
-  line-height: 3.75rem;
+  line-height: .9;
```
SamuelRandall: Should this have a unit of measurement attached to it?

styleguide/derek/_partials/solutions/header-smallIMAC.ejs
```diff
@@ -0,0 +1,18 @@
+<!--Header Short-->
+<div class="imacBanner small">
+  <div class="imacBanner-container">
+    <div class="imacBanner-textRow">
+      <div class="col-sm-5 imacBanner-textContainer">
```
SamuelRandall: Can we combine the `col-sm-5` into the `imacBanner-textContainer` class?

styleguide/derek/_partials/solutions/header-largeIMAC.ejs
```diff
@@ -1,18 +1,20 @@
...
-      <div class="col-sm-7">
-        <div class="imacBanner-text">
-          <h2 class="imacBanner-headline">3,000 cloud experts working for you 24/7/365?<br/>
-          Sounds like you may need a bigger office.</h2>
-        </div>
+<!--Header Long-->
+<div class="imacBanner large">
+  <div class="imacBanner-container">
+    <div class="imacBanner-textRow">
+      <div class="col-sm-7 imacBanner-textContainer">
```
SamuelRandall: Seems a bit odd to be using the bootstrap class when everything else is maintainable. This may be a situation where `imacBanner-textContainer` should be a unique class for each size banner.

styleguide/derek/_partials/solutions/header-mediumIMAC.ejs
```diff
@@ -0,0 +1,22 @@
...
+    <div class="imacBanner-textRow">
+      <div class="col-sm-6 imacBanner-textContainer">
+        <h2 class="imacBanner-headline imacBanner-headlineMedium">the cloud is a mystery wrapped in a riddle.<br/>We can unwrap it for you</h2>
+        <span class="line left-top"></span>
+        <span class="line left-bottom"></span>
+        <span class="line right-bottom"></span>
+      </div>
+    </div>
+    <div class="imacBanner-buttonRow">
+      <a class="button imacBanner-button">Read Our Case Studies  <i class="fa fa-angle-right imacBanner-buttonIcon"></i></a>
```
SamuelRandall: Can we copy the styles from `button` into `imacBanner-button` so avoid using 2 classes here?

styleguide/_themes/global/scss/banners.scss
amberRrucker: Nope! Since fit text dynamically defines the text size it will just be 90% of whatever fit text defines it to be. It works like a champ.

styleguide/derek/_partials/solutions/header-largeIMAC.ejs
amberRrucker: I'm leaving these until I have a discussion with brand to make sure these work for them and they're all approved.

styleguide/derek/examples/imac.jade
amberRrucker: The example banners hadn't yet been created when those are called within their ejs files so yes. And the ejs files are included on other pages so should have those calls within them to be modular. As soon as we're all design final (this afternoon) we will have to discuss how we want them included into the www application of this.

styleguide/derek/_partials/solutions/header-largeIMAC.ejs
amberRrucker: I was trying to avoid adding an additional class if possible but we can add that in

styleguide/_themes/global/scss/vendor_prefixes.scss
amberRrucker: Good call

styleguide/_themes/global/scss/banners.scss
amberRrucker: ill make that change if they end up actually having the same params for fittext.

WidgetsBurritos:
image

Should the lines go all the way down on the left in larger resolutions?

styleguide/_themes/global/scss/banners.scss
WidgetsBurritos: If you're gonna base it on percentages, then I'm wondering if it should be `90%` instead of `.9` . Having no unit just reads weird, whereas `90%` has no ambiguity.

WidgetsBurritos: Or actually on second thought, `line-height: .9em;` has the exact same effect and is more consistent with what we're wanting. `em` is always relative to the closest font-size, which is being set a few lines above, so this will always be 90% of the font size.

styleguide/derek/_partials/solutions/header-largeIMAC.ejs
WidgetsBurritos: I'd say go ahead and add the extra class.

amberRrucker:
Alrighty @WidgetsBurritos i think i tackled everything in here we talked about yesterday.

WidgetsBurritos:
@amberRrucker see comments above about line-height.  That's still unresolved.

styleguide/_layouts/default.jade
```diff
@@ -17,9 +17,23 @@ html(lang="en")
     link(href="#{ css }", rel="stylesheet")
     link(href="css/main.css", rel="stylesheet")
     link(href="https://fonts.googleapis.com/css?family=Open+Sans:700,300,200,600,800,400|Khand:700,600,500|Montserrat:400,700", rel="stylesheet")
+    script(type="text/javascript").
+      WebFontConfig = {
+        google: { families: [ 'Khand:700' ] }
+      };
```
WidgetsBurritos: If we're going to do this, we probably should use this to pull in all google font families, not just Khand:700.  At this point, we're double loading the font.

styleguide/_themes/global/scss/banners.scss
```diff
@@ -36,31 +36,147 @@ $banner-text-margin-medium: 10%;
...
+  @include make-container;
+  @include fade(.25s, ease-in-out);
+  opacity: 0;
+}
+
+.wf-khand-n7-active {
+  .imacBanner-container {
+    opacity: 1;
+  }
+}
```
WidgetsBurritos: So if we're going to do this, which I am in support of by the way, I think it makes sense to have all of these webfont overrides in their own file that are excluded from sass linting.  The main reason for that is this class names don't meet sass-lint guidelines, but since they're mandated by an external tool, I don't think we should have that causing issues.  
```
styleguide/_themes/global/scss/banners.scss
  74:2  warning  Class '.wf-khand-n7-active' should match regular expression /^([a-z0-9]+([A-Z][a-z0-9]+)*)(-{1}([a-z0-9]+([A-Z][a-z0-9]+)*)){0,2}$/  class-name-format
```

Alternatively, we can change our class name rules to allow 4 tokens, but I'm not sure we want to keep adjusting our rules everytime we come up with an exception.  That said, excluding from the rules may not be a good solution either.  This probably warrants further discussion with the team.

styleguide/_themes/derek/scss/snowflakes/homepage.scss
```diff
@@ -18,15 +18,15 @@
...
+  p {
     letter-spacing: 1px;
     line-height: 24px;
     font-size: 16px;
     padding-bottom: 10px;
   }
 }
 
-#hosting-options{
+#hosting-options {
```
WidgetsBurritos: I'm all about cleaning sass lint errors, but if it's unrelated to this PR, not sure it's worth the effort.  But if we do decide we're okay with small incremental fixes like this, then we need to make sure the `package.json` file gets updated to reference the new error count. 

This more seems like an trade-off to allow one new error into the system by getting rid of two.  No new code should introduce new code to sass lint. If it does we need to figure evaluate if we want to exclude, or modify the rules instead.

styleguide/_themes/global/scss/vendor_prefixes.scss
```diff
@@ -52,3 +58,13 @@
   -webkit-filter: sepia(.9);
   filter: sepia(90%);
 }
+
+@mixin text-stroke($width, $color) {
+  -webkit-text-stroke: $width $color;
+}
+
+@mixin fade($duration, $type) {
```
WidgetsBurritos: While I definitely like having a generic mixin here to specify a duration and easing function, I'm wondering if it makes sense to just set the values you mentioned separately as a default `$duration` and `$type` value for this mixin.  I think this can help avoid inconsistencies in the future, and will allow us to just do `@include fade;` wherever we need this effect.

styleguide/_themes/global/scss/banners.scss
```diff
@@ -36,31 +36,147 @@ $banner-text-margin-medium: 10%;
...
+  @include make-sm-column(6);
+}
+
+.imacBanner-headlineSmall {
+  @include make-sm-column(5);
+}
+
 .imacBanner-headline {
   @include heading($white);
+  font-family: 'Khand';
```
WidgetsBurritos: It probably makes sense to define this font as a `$imac-font` variable or something like that, and place it in `styleguide/_themes/global/scss/fonts.scss` instead of just explicitly writing it out here.

styleguide/_themes/global/scss/banners.scss
```diff
@@ -36,31 +36,147 @@ $banner-text-margin-medium: 10%;
...
   font-weight: 700;
-  line-height: 3.75rem;
+  line-height: .9;
   margin: 0;
 }
 
-.imacBanner-text {
+.imacBanner-textContainer {
   border: 3px solid $white;
+  margin: auto;
```
WidgetsBurritos: Does it make sense to have `auto` on the top/bottom margins here as well? Would `0 auto` be a preferable value?

styleguide/_themes/global/scss/buttons.scss
```diff
@@ -6,10 +6,10 @@
   display: inline-block;
   font-size: .85em;
   font-weight: 600;
-  padding: 15px 20px;
+  padding: 8px 10px;
   text-align: center;
   text-decoration: none;
-  text-transform: uppercase;;
+  text-transform: uppercase;
```
WidgetsBurritos: But that semi-colon is gonna be lonely now!

styleguide/derek/_partials/solutions/header-mediumIMAC.ejs
```diff
@@ -0,0 +1,17 @@
+<!--Header Medium-->
+<div class="imacBanner medium">
+  <div class="imacBanner-container">
+    <div class="imacBanner-textRow">
+      <div class="imacBanner-textContainer imacBanner-headlineMedium">
+        <h2 class="imacBanner-headline">the cloud is a mystery wrapped in a riddle.<br/>We can unwrap it for you</h2>
```
WidgetsBurritos: These `<br />` tags can make the text wrapping look a little strange, depending on your browser size. Is this a requirement here?

styleguide/derek/_partials/solutions/header-smallIMAC.ejs
```diff
@@ -0,0 +1,13 @@
+<!--Header Short-->
+<div class="imacBanner small">
+  <div class="imacBanner-container">
+    <div class="imacBanner-textRow">
+      <div class="imacBanner-textContainer imacBanner-headlineSmall">
+        <h2 class="imacBanner-headline">the cloud runs on its own,<br/>with lots of help</h2>
```
WidgetsBurritos: Should these have ending punctuation and start with capital letters? 

Yes I know the text-transform does that already, but It just looks cleaner if the grammar is proper. I think if people are using this as their point of reference for our branding styles, it shouldn't create the impression that grammar is unimportant, but that's just my two cents.

styleguide/derek/examples/imac.jade
```diff
@@ -0,0 +1,198 @@
...
+    .col-md-12
+      h2.red IMAC Banner Examples and Usage
+
+  .row
+    .col-md-12
+      h5.red Banners
+    .col-md-8
+      p Our banners are the leading graphic element on all pages and therefore one of our largest vehicles for our brand on the page.
+      p These banner types should be used on IMAC campaign pages. A CTA option is included within all banners sections, but is not necessary to use.
+      p Addition Photographic Requirements for Web Banners:
```
WidgetsBurritos: Additional*

styleguide/derek/examples/imac.jade
```diff
@@ -0,0 +1,198 @@
...
+
+include ../_partials/solutions/header-smallIMAC.ejs
+
+hr
+
+.container
+  .row.half-padding-full
+    .col-md-12
+      h2 Examples
+      p All the following are examples of headlines given by Y&R and how they look/will be put into banner areas. If we need to remove/add buttons and button text we can make that happen. There were no examples of internal pages. Still need photography examples.
```
WidgetsBurritos: Ampersands should be referenced as `&amp;`

styleguide/js/jquery.fittext.js
```diff
@@ -0,0 +1,43 @@
...
+      resizer();
+
+      // Call on resize. Opera debounces their resize by default.
+      $(window).on('resize.fittext orientationchange.fittext', resizer);
+
+    });
+
+  };
+
+})( jQuery );
```
WidgetsBurritos: I still don't see any evidence on my system that fit text is actually doing anything.  I've removed it from my local branch, and I see zero difference between the two.

WidgetsBurritos: Nevermind that last item.  I ran a comparison between the two in backdrop, and can see the differences now .

styleguide/_themes/derek/scss/snowflakes/homepage.scss
amberRrucker: had to update it because of the sass-lint error from the khand load stuff

styleguide/_themes/global/scss/banners.scss
amberRrucker: left note in code about this one.

styleguide/derek/_partials/solutions/header-mediumIMAC.ejs
amberRrucker: Its a requirement from the campaign, since the lines are two separate thoughts.

styleguide/derek/_partials/solutions/header-smallIMAC.ejs
amberRrucker: Fixed the captilization but the punctuation has been brought to brands attention so we will see. For now this is what I was given for headlines unfortunately.

styleguide/js/jquery.fittext.js
amberRrucker: Very odd. Mine is working as expected and when i run it through browserstack its working as expected there as well. Is anyone else experiencing this problem?

styleguide/_themes/global/scss/banners.scss
amberRrucker: lemme know where you want it and i will put it in that place.

WidgetsBurritos:
@amberRrucker I'm adding WIP to this ticket until you get those keyframe changes we discussed yesterday in here.

amberRrucker:
Done and Ready. I cant move this to QE because I forgot my RSA thing. @SyntaxStacks I love you?

WidgetsBurritos:
+1

SyntaxStacks:
QE +1

styleguide/_layouts/default.jade
```diff
@@ -20,6 +20,7 @@ html(lang="en")
     link(href="https://maxcdn.bootstrapcdn.com/font-awesome/4.6.1/css/font-awesome.min.css" rel="stylesheet")
     script(src="https://code.jquery.com/jquery-1.12.3.min.js" integrity="sha256-aaODHAgvwQW1bFOGXMeX+pC4PZIPsvn2h1sArYOhgXQ=" crossorigin="anonymous")
     script(src="https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/zoolander/modernizr.min.js")
+    script(src="#{base}js/jquery.fittext.js")
```
WidgetsBurritos: @amberRrucker While this is still open, can you swap this to use the CDN and kill the file from the repo? 
`https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/zoolander/jquery.fittext.js`

amberRrucker:
@WidgetsBurritos I removed fittext and included the cdn link

styleguide/_layouts/default.jade
```diff
@@ -20,6 +20,7 @@ html(lang="en")
     link(href="https://maxcdn.bootstrapcdn.com/font-awesome/4.6.1/css/font-awesome.min.css" rel="stylesheet")
     script(src="https://code.jquery.com/jquery-1.12.3.min.js" integrity="sha256-aaODHAgvwQW1bFOGXMeX+pC4PZIPsvn2h1sArYOhgXQ=" crossorigin="anonymous")
     script(src="https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/zoolander/modernizr.min.js")
+    script(src="`https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/zoolander/jquery.fittext.js")
```
WidgetsBurritos: Did you confirm this works, because I'm pretty sure that backtick there is gonna be an issue.

WidgetsBurritos: Yup.  Definitely an issue. 
image

amberRrucker:
@WidgetsBurritos whoooooops fixed it!

WidgetsBurritos:
@amberRrucker No prob.  

+1 on the cdn update.

PR #211, opened 2016-08-09
RSWEB-6910 Fix Hover State grid pattern

SamuelRandall:
Jira: https://jira.rax.io/browse/RSWEB-6910

This includes a WIP refactor of the keyoffers and business resources grid patterns.

PR #213, opened 2016-08-11
RSWEB-6984: Add mini-panel pattern support in Zoolander

WidgetsBurritos:
Jira: https://jira.rax.io/browse/RSWEB-6984

This PR adds mini-panel support for patterns into zoolander. 

This PR actually pulls in another commit made by @SamuelRandall in #211 in addition to my own changes. That PR was already +1'd multiple times, but we couldn't merge it due to dependency issues. 

The main concept behind the changes made here is that to get this to work, we had to convert pattern `container` classes into `container-fluid`, as well as the maintainable css mixins that correspond to those classes. We also moved the 960px constraint from the containers to the rows.  As such, this required creating an alternative mixin to `make-row`, which I have called `rs-make-row`.  This has to be used in place of `make-row` moving forward unless we remove width constraints. 

I will leave further explanation to various components in the code below.


styleguide/_themes/global/scss/grid.scss
```diff
@@ -1,10 +1,20 @@
+$max-row-width: 960px;
+
+@mixin rs-make-row($gutter: $grid-gutter-width) {
+  @include make-row($gutter);
+  margin: 0 auto;
+  max-width: $max-row-width;
+}
```
WidgetsBurritos: Unfortunately I couldn't just write over the existing `make-row` mixin, so I had to create an alternative, that basically extends basic `make-row` functionality, but also centers it, and sets a maximum width. All existing references to `make-row` have thus been changed to `rs-make-row`

styleguide/_themes/derek/scss/components/ticker.scss
```diff
@@ -17,11 +17,11 @@ $ticker-span-bg: rgba(200, 0, 0, .5);
 }
 
 .ticker-container {
-  @include make-container;
+  @include make-container-fluid;
```
WidgetsBurritos: As illustrated here, the general idea across the board is that all containers should be swapped to container-fluid. With this being such a sweeping change, I imagine some heavy QE will be necessary for this to make sure nothing else breaks in the process.

styleguide/_themes/derek/scss/patterns/cta.scss
```diff
@@ -0,0 +1,68 @@
...
+  text-align: center;
+}
+
+// Subpanels.
+.ctaSubpanel-columnButton {
+  @include make-sm-column(10);
+  @include make-sm-column-push(1);
+  @include make-md-column(10);
+  @include make-md-column-push(1);
+}
```
WidgetsBurritos: As illustrated here, subpanel-specific behavior may require additional classes to be created. In the case of the CTAs, the buttons were extremely narrow inside subpanels so we had to expand the width.  CTAs also featured a rewrite to maintainable css, but I tried to keep all the other patterns intact as much as possible.

styleguide/_themes/global/scss/grid.scss
```diff
@@ -16,3 +26,12 @@
     width: $container-lg;
   }
 }
+
+.container,
+.container-fluid {
+  position: relative;
```
WidgetsBurritos: So I'm a little nervous about adding this `position: relative` here. I don't think it's gonna affect anything, unless we're using absolute positioning relative to the document root anywhere.  The purpose of it, was to give me an anchor point for our anchored text behavior in zoolander, but I guess we probably need to do some thorough testing of www.

styleguide/css/main.scss
```diff
@@ -1,13 +1,9 @@
-$white: #fff;
-$light-gray-white: #efefef;
-$dark-gray: #333;
-$medium-gray: #666;
-
+@import '../_themes/global/scss/vendor_prefixes';
+@import '../_themes/global/scss/swatch';
```
WidgetsBurritos: I pulled in the color swatch and vendor prefixes into main.scss just so we don't have to maintain separate lists for these color swatches. Since these includes are mixins/variables only, they should have zero effect on the final contents of main.css

styleguide/css/main.scss
```diff
@@ -47,9 +43,50 @@ $medium-gray: #666;
...
 .openbtn {
-  background-color: $dark-gray;
+  background-color: $gray-dark;
   left: 0;
   padding: 10px;
   position: absolute;
   top: 0;
 }
+
+.anchored-text {
```
WidgetsBurritos: Nesting panels created an issue where the help text for a pattern was actually getting in the way and creating a misrepresentation of what the mini panel would look like. As a result I opted to just create little icons that people can click on to get more details about the pattern, and that gets placed inside of `anchored-text`.  Since this is zoolander-only behavior, I placed it in main.scss to avoid adding unnecessary code to the codebase.

styleguide/derek/_partials/solutions/additional-resources.ejs
```diff
@@ -1,48 +1,70 @@
+<%
+// While seemingly redundant, this is actually necessary for EJS to evaluate variable scope.
+var isSubpanel = !!isSubpanel;
```
WidgetsBurritos: So apparently ejs is particular about variable scope, which makes this line necessary. Without it, the `isSubpanel` value wasn't carrying through. 

As far as an explanation as to what this is, this is basically how we can identify if we are in a subpanel or not. This allows us to reuse as much of the existing patterns as possible, and just create exceptions for subpanels instead of having two separate bits of code to maintain for ever single pattern.

styleguide/derek/examples/subpanels.ejs
```diff
@@ -0,0 +1,158 @@
...
+  'services-5-col': {'title': 'IAS: 5 cols'},
+  'services-6-col': {'title': 'IAS: 6 cols'},
+  'key-offers': {'title': 'Key Offers'},
+  'lead': {'title': 'Lead'},
+  'list': {'title': 'List'},
+  'testimonials': {'title': 'Quote'},
+  'text': {'title': 'Text'},
+  'value-prop': {'title': 'Value Prop'}
+};
+%>
```
WidgetsBurritos: The way this page works is that it defines the layouts/patterns as simple JS objects. We then build out the HTML based on this info.

styleguide/derek/examples/subpanels.ejs
```diff
@@ -0,0 +1,158 @@
...
+        <option value="<%- pattern %>">[Right] <%- patterns[pattern].title %></option>
+        <% } %>
+      </select>
+    </div>
+    <div class="col-sm-4">
+      <input type="submit" class="btn btn-xs btn-default" id="search-mini-panels" value="Go to Panel"/>
+      <input type="reset" class="btn btn-xs btn-info" id="reset-mini-panels" value="Back to Top"/>
+    </div>
+  </div>
+</div>
```
WidgetsBurritos: This pattern selector provides a way for users to select a specific layout/left pattern/right pattern.  There is corresponding javascript for this at the bottom of the file.

styleguide/derek/examples/subpanels.ejs
```diff
@@ -0,0 +1,158 @@
...
+      var layout = $layout.val();
+      var pattern_left = $pattern_left.val();
+      var pattern_right = $pattern_right.val();
+      var id = '#pattern_' + layout + '_' + pattern_left + '_' + pattern_right;
+
+      $body_html.animate({scrollTop: $(id).offset().top - selectorHeight}, 250);
+    });
+  });
+})(jQuery);
+</script>
```
WidgetsBurritos: Since this is file-specific JS behavior, I figured leaving it here was probably okay.

styleguide/derek/examples/subpanels.ejs
```diff
@@ -0,0 +1,158 @@
...
+      <%- partial(file_right, {isSubpanel: true}) %>
+    </div>
+  </div>
+</div>
+<hr/>
+<%
+    }
+  }
+}
+%>
```
WidgetsBurritos: The above essentially creates every combination of layout, left pattern and right pattern and includes the necessary files, all while setting `isSubpanel` to `true`, so we can see our pattern alternatives.

styleguide/js/zoolander.js
```diff
@@ -77,6 +77,17 @@ jQuery(document).ready(function($) {
...
+  // Hides/shows anchored text within zoolander (zoolander-only).
+  $('.anchored-text').click(function() {
+      var isVisible = $(this).hasClass('visible');
+      if (isVisible) {
+          $(this).removeClass('visible');
+      }
+      else {
+          $(this).addClass('visible');
+      }
+  });
```
WidgetsBurritos: Very simple javascript to make the anchored-text behavior work.

styleguide/_themes/derek/scss/components/regionSelector.scss
```diff
@@ -105,7 +105,7 @@
 }
 
 .regionSelector-row {
-  @include make-row;
+  @include rs-make-row;
```
amberRrucker: ?

styleguide/_themes/derek/scss/components/ticker.scss
amberRrucker: how are we going fluid?

styleguide/_themes/derek/scss/components/regionSelector.scss
WidgetsBurritos: See my explanation in my comments above.

styleguide/_themes/derek/scss/patterns/cta.scss
```diff
@@ -0,0 +1,67 @@
...
+.ctaSubpanel-columnSecondary,
+.ctaSubpanel-columnTitle {
+  @include make-md-column(12);
+  @include make-md-column-push(0);
+}
+
+.ctaSubpanel-columnIcon {
+  @include make-xs-column(12);
+  @include make-xs-column-push(0);
+
```
amberRrucker: you got some extra space here

styleguide/_themes/derek/scss/components/ticker.scss
WidgetsBurritos: i explained this elsewhere, but basically we're forcing width restrictions on row, instead of container.

styleguide/_themes/derek/scss/patterns/cta.scss
```diff
@@ -0,0 +1,67 @@
...
+  text-align: center;
+  width: 100%;
+}
+
+.ctaSubpanel-columnIconImage {
+  max-width: 70px;
```
amberRrucker: I dont love this. Can we not have it set by its container?

styleguide/_themes/derek/scss/patterns/grid/keyOffers.scss
```diff
@@ -1,35 +1,69 @@
...
-    }
-
-    p {
-      bottom: 0;
-      float: left;
-      left: 25px;
-      padding-bottom: 0;
-      position: absolute;
-    }
-  }
```
amberRrucker: This is not the only place "bottom" is used. There's things on www utilizing this I think. We may need a very solid QE sweep around this one.

styleguide/_themes/global/scss/fonts.scss
```diff
@@ -20,7 +21,7 @@ $code: Tahoma, Arial, Geneva, Helvetica;
 }
 
 @mixin copy($color: $gray-darkest) {
-  color: $gray-darkest;
+  color: $color;
```
amberRrucker: copy should never be any other color. If we want to do this we should define a $brand-copy or $font-base variable.

styleguide/_themes/derek/scss/patterns/grid/keyOffers.scss
SamuelRandall: There is another `.bottom` class in modifiers. This was solely for the keyoffers pattern but that one still exists.

styleguide/_themes/global/scss/grid.scss
amberRrucker: no alternatives huh? that sucks. we will have to make sure this is all good in browerstack. Will this effect any of the general content types that utilize containers and rows now?

styleguide/_themes/global/scss/grid.scss
amberRrucker: I'm also nervous. A good bit of our general content has containers and container-fluid in them.

styleguide/css/main.scss
amberRrucker: thank you sir.

styleguide/css/main.scss
```diff
@@ -47,9 +43,50 @@ $medium-gray: #666;
...
+  @include border-radius(1.5rem);
+  left: 0;
+  padding: 1.5rem;
+  position: absolute;
+  text-indent: -100000px;
+  top: 0;
+  z-index: 99999;
```
amberRrucker: 0_____0 we can probably have a treatment that doesnt need so much styling.

amberRrucker:
I'll pull this down and try and find a more light weight way to display the anchored text labels. Sucks how much code we had to add for subpanel/panel conditional stuff but I trust this was the best way.

styleguide/css/main.scss
WidgetsBurritos: yeah I just through some stuff together.  we can adjust the styling.  no biggie.

WidgetsBurritos:
@amberRrucker Yeah I'm not totally thrilled with this solution either.  But it does _"solve"_ the problem we're having. I'm definitely down for suggestions, and minimizing some of the anchoring stuff.  I was just kinda finding a temporary solution and figured we could massage it as we saw fit.

styleguide/_themes/global/scss/grid.scss
WidgetsBurritos: well in theory we're just shifting the 960 restriction so it *should* work.  Should is the key word.  I think this is gonna need a heavy hand as far as QE goes, and probably is gonna extend into next sprint at this point.

styleguide/_themes/global/scss/fonts.scss
WidgetsBurritos: I vaguely remember there being a reason for this, but can't recall what it is now.

styleguide/_themes/derek/scss/patterns/cta.scss
bighappyface: +1

Setting width to 100% and then setting the max width to 70px is... odd.

bighappyface:
Considering the work to maintain 960 what is preventing us from going full fluid? It is an inevitable change that we will make and this kind of work may not make that transition easier.

amberRrucker:
@bighappyface we're held back by any legacy code in our general content types at this point - which would be a problem with this work anyways since none of that has any row declarations. I would rather we just unique rules based on if its in a subpanel or not, not globally. It may also prevent people from using the styleguide since we're sort of creating our own version of bootstrap at this point. I think maybe a larger discussion around possible fixes may be in order?

styleguide/_themes/global/scss/fonts.scss
WidgetsBurritos: So after thinking on this more, this was actually behavior we had added in to the `heading()` and `subhead()` mixins for other reasons, and I think just for consistency I added it to `copy()` here.  We're not actually setting the color anywhere, but it's intact as a just-in-case method, but we could just scrap it if you'd prefer.

styleguide/_themes/derek/scss/patterns/grid/keyOffers.scss
WidgetsBurritos: What Sam said.

WidgetsBurritos:
@amberRrucker @bighappyface Okay, I've refactored with that strategy in mind.  Have another look

amberRrucker:
Much better. I'm okay with this. I'm pretty sure with this solution we shouldn't have any trouble with our general content but I still think we should allow for some significant QE time probably. +1

styleguide/_themes/derek/scss/patterns/lead.scss
```diff
@@ -21,3 +26,10 @@
   text-align: center;
 }
 
+@media only screen and (min-width: $screen-sm-min) {
+  width: 25%;
```
bighappyface: Where is the selector this property applies to?

styleguide/_themes/derek/scss/patterns/lead.scss
```diff
@@ -21,3 +26,10 @@
...
+}
+
+@media only screen and (min-width: $screen-md-min) {
+  width: 16.7%;
```
bighappyface: Where is the selector this property applies to?

WidgetsBurritos: derp.  I'll fix that.

styleguide/derek/_partials/solutions/testimonials.ejs
```diff
@@ -1,26 +1,47 @@
+<%
+// While seemingly redundant, this is actually necessary for EJS to evaluate variable scope.
+var isSubpanel = !!isSubpanel;
```
bighappyface: Considering this snippet:

```ejs
<%- partial(file_left, {isSubpanel: true}) %>
```

Does the value of `isSubpanel` come in as the string `'true'` or the boolean `true`?

styleguide/_themes/derek/scss/patterns/grid/businessResource.scss
```diff
@@ -1,14 +1,24 @@
-.resource {
+$hover-color: 12px solid rgba(0, 0, 0, .4);
```
bighappyface: Could this be used elsewhere? If not, why not set the property directly and skip the variable?

styleguide/_themes/derek/scss/patterns/grid/productFeatures.scss
```diff
@@ -30,6 +31,17 @@
   }
 }
 
+.subpanel {
+  .features-container {
+    @include make-md-column(12);
+    margin: 5px 0;
+
+    .rsweb {
```
bighappyface: `.subpanel .features-container .rsweb` :\

styleguide/_themes/global/scss/fonts.scss
```diff
@@ -2,6 +2,7 @@ $heading: 'Open Sans', Helvetica, Arial, sans-serif;
 $subhead: 'Montserrat', Helvetica, Arial, sans-serif;
 $copy: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
 $code: Tahoma, Arial, Geneva, Helvetica;
+$semi-bold: 600;
```
bighappyface: Does this weight apply to semi-bold for all fonts?

ejchiefs:
Just went through a visual review with @WidgetsBurritos. While there are some visual issues with icons on some of the patterns on 66/33, we can communicate that to the ECMs to make them aware. They can either avoid specific patterns in the corner pocket or avoid using icons in those patterns when in the corner pocket. We also need to remove the bg-light-gray class since these patterns will be placed within rivers unless specified in the admin. 

👍  to move forward from an aesthetics perspective

styleguide/_themes/global/scss/fonts.scss
WidgetsBurritos: Ehh,  I'll probably scrap it.  I was originally planning on creating a variable that could be used across the board, but that probably should be done in another task, because there are other weights to take into consideration as well.

styleguide/derek/_partials/solutions/testimonials.ejs
WidgetsBurritos: A boolean.  I confirmed it in the ejs file.

styleguide/_themes/derek/scss/patterns/grid/businessResource.scss
WidgetsBurritos: That came in via Sam's changes in PR #211. I didn't catch it in the code review, but that's probably unnecessary so I'll just pull it out here.

styleguide/_themes/derek/scss/patterns/grid/productFeatures.scss
WidgetsBurritos: We probably need a follow up task to refactor the features as they aren't using maintainable css. Until that happens, we need a way to distinguish between `.rsweb` inside of features, vs inside of other patterns, inside of subpanels. I'm not a fan of this, by any means, and I guess could just refactor this one portion if you'd prefer, but given that this will need follow up attention anyway, I'm wondering if it makes sense to just clean it up all at once?

styleguide/derek/_partials/solutions/testimonials.ejs
bighappyface: Then the reason this is here is to make up for the fact that `isSubpanel` is undefined in most cases.

If you are so inclined you can set it globally to `false` in the `_harp.json` globals:

https://github.com/rackerlabs/zoolander/blob/c04259429f2fdb5e18b4055626d0f4371c90d7af/styleguide/_harp.json#L2

bighappyface:
+1

WidgetsBurritos:
Since we're rethinking this strategy, I'm closing this PR for now.

PR #214, opened 2016-08-15
feat(landing-page):adjusting global elements

amberRrucker:
https://jira.rax.io/browse/RSWEB-6972
Adding landing page version of footer and adjusting ceiling to include title.

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -15,5 +15,5 @@ rules:
     - convention-explanation: 'Classes should be camelCase and separated by no more than 3 hyphens'
   mixin-name-format:
     - 1
-    - convention: ^([a-z0-9]+([A-Z][a-z0-9]+)*)(-{1}([a-z0-9]+([A-Z][a-z0-9]+)*)){0,5}$
-    - convention-explanation: 'Classes should be camelCase and separated by no more than 5 hyphens'
+    - convention: ^([a-z0-9]+([A-Z][a-z0-9]+)*)(-{1}([a-z0-9]+([A-Z][a-z0-9]+)*)){0,4}$
+    - convention-explanation: 'Mixins should be camelCase and separated by no more than 4 hyphens'
--- a/styleguide/_themes/derek/scss/globalElements/basement.scss
+++ b/styleguide/_themes/derek/scss/globalElements/basement.scss
@@ -5,6 +5,15 @@
   }
 }
 
+//Social Icon Color Mixin
+@mixin basement-beforeColor($class, $color) {
+  #{$class} {
+    &::before {
+      color: $color;
+    }
+  }
+}
+
 // Basement Classes
 .basement {
   padding: 10px 15px;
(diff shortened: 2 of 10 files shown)


.sass-lint.yml
```diff
@@ -11,7 +11,7 @@ rules:
   no-url-protocols: 0
   class-name-format:
     - 1
-    - convention: ^([a-z0-9]+([A-Z][a-z0-9]+)*)(-{1}([a-z0-9]+([A-Z][a-z0-9]+)*)){0,2}$
+    - convention: ^([a-z0-9]+([A-Z][a-z0-9]+)*)(-{1}([a-z0-9]+([A-Z][a-z0-9]+)*)){0,4}$
     - convention-explanation: 'Classes should be camelCase and separated by no more than 5 hyphens'
```
WidgetsBurritos: Does this need to be 4, or will 3 suffice?  I don't see any rules below that warrant support for 4 hyphens. 

Additionally, the explanation should match whatever value we select. 

Also, we probably should have a team discussion about this rule in general.  If we're having to keep changing things because of unavoidable third-party dependency, then maybe we should discuss the merits of this rule in general, because we may be defeating the purpose of this rule by setting a threshold so high to begin with.

WidgetsBurritos:
There is a faint, but noticeable white edge around the radius on the social icons.  It looks a little pixelated to me. 

image

To help better illustrate, here's the `fa-stack-1x` class set to yellow instead of white, to help accentuate the border issue: 

image

And zoomed in a little bit: 
image

amberRrucker:
weird that it looks pixelated since its an svg from font awesome but ill look into it.

.sass-lint.yml
amberRrucker: taking it to 5 reduced our error count from 109 to 106. But i agree i think we should have a larger discussion around this rule.

amberRrucker:
screen shot 2016-08-17 at 9 28 45 am
So i zoomed in to 300% and got out my magnifying glass and I'm kind of barely seeing it. Can anyone else duplicate the pizelation though? I can't duplicate that at all.
@WidgetsBurritos

amberRrucker:
screen shot 2016-08-17 at 9 50 28 am
@WidgetsBurritos I made some changes in order to get this good to go but I couldnt reeeeeally see the problem before. I attached a screen shot of the inner square now so it couldnt possibly be making a border so hopefully that fixes it if you could confirm?

WidgetsBurritos:
@amberRrucker that seems to have fixed it.

styleguide/_themes/derek/scss/globalElements/basement.scss
```diff
@@ -83,3 +89,59 @@
...
+}
+
+.basement-iconListItem {
+  float: left;
+}
+
+.basement-iconBg {
+  color: $white;
+  display: inline-block;
+  font: normal normal normal 1.83em FontAwesome;
```
WidgetsBurritos: Why is normal set three times here?

amberRrucker: I had to copy over some stuff from font awesome for the 'fa' class since it was overriding font size. It looks like those were unnecessary so I removed them.

.sass-lint.yml
WidgetsBurritos: @amberRrucker Spoke with @SyntaxStacks about this issue. Can we create a class called `.faBefore-white`, or something like that and just add it to these references in the footer?   That will get rid of the this issue in this context.  We'll still have to solve it elsewhere, but that at least helps us not have it hold up this PR.

amberRrucker: @WidgetsBurritos and @SyntaxStacks this is not to solve for the white square stuff. This is to solve for applying a color to the google plus icon.

WidgetsBurritos: @amberRrucker Sorry I referenced the wrong thing, but what we're suggesting is something like this: 

Instead of doing this:
```
.fa-google-plus-square {
   &::before {
     color: $googleplus-swatch;
   }
 }
```

Do this:

```
.faBefore-googleplus {
   &::before {
     color: $googleplus-swatch;
   }
 }
```

And then add the class to `<i class="fa fa-2x fa-google-plus-square fa-stack-2x"></i>`

And for consistency, just do that for all the classes.

WidgetsBurritos: That way, we're not competing with font awesome at all.

WidgetsBurritos:
👍  -- code review

@runrog can you give this a design review?

runrog:
@amberRrucker @WidgetsBurritos Design Review: So on mobile the social media icons aren't clickable, it looks like some portions of the icons are but it's behaving janky. 

On all viewports, the top logo has too much white space underneath it, not sure if that is intentional but it just looks off. 

On mobile, (around 760px and <) the top title ("RACKSPACE® PRIVATE CLOUD POWERED BY OPENSTACK®") disappears.

landing-page-footer
landing-page-top

amberRrucker:
@runrog the banner hasnt been replaced in this PR so the whitespace is there because it gets fixed in that PR. The title disappears on mobile on purpose. I'll check out the social icon clickness.

amberRrucker:
@SyntaxStacks I've rebased and solved the click issue @runrog found. I know I'll need to rebase again once the other landing page branch gets merged though so let me know when that happens or which one is going first and ill rebase accordingly. :)

runrog:
👍

PR #216, opened 2016-08-17
bug(mobile):subnav fixes

amberRrucker:
https://jira.rax.io/browse/RSWEB-7047

Subnav reformat and mobile fix can be seen here:
http://localhost:9000/derek/examples/office365

Looking for better mobile performance and issue fixes around these areas:
- incorrect breakpoints
- page title display issues in mobile
- exploring another active state for mobile menu

There has also been pretty significant reformatting of the code which will need a www PR for implementation.

### DIFF
--- a/styleguide/_themes/derek/scss/components/subnav.scss
+++ b/styleguide/_themes/derek/scss/components/subnav.scss
@@ -1,91 +1,160 @@
-//subnav
 .subnav {
-  background-color: $blue-gray;
+  background-color: $gray-dark;
   border-radius: 0;
+}
 
-  .nav > li > a:focus {
-    background-color: transparent;
-  }
+.subnav-container {
+  @include make-container;
+}
 
-  .nav > li.no-hover > a:focus {
-    background-color: $brand-success;
-  }
+.subnav-links {
+  list-style-type: none;
+  padding: 0;
+}
 
-  .subnav-inside {
-    width: 100%;
-  }
+.subnav-header {
+  float: left;
+  width: $full-width;
+}
 
-  .navbar-header {
+.subnav-collapseBtn {
+  @include make-xs-column(1);
+  padding-left: 0;
+}
 
-    .col-xs-10 {
-      margin-top: 10px;
-    }
+.subnav-toggleBtn {
+  float: left;
+  margin: 14px 0;
+  padding: 0;
 
-    button {
-      background-color: transparent;
-    }
-
-  }
-
-  .button {
-    margin: 7px 0;
-  }
-
-  .icon-bar {
-    background-color: $white;
-  }
-
-  li {
-    border-bottom: solid transparent 4px;
-    float: left;
-    padding: 7px 7px 6px;
-
-    &:hover {
-      background-color: transparent;
-      border-bottom: solid $teal-dark 4px;
-    }
-
-    a {
-      color: $white;
-      font-weight: 400;
-      text-decoration: none;
-
-      &:hover {
-        background-color: transparent;
-        text-decoration: none;
-      }
-    }
-
-    &.no-hover {
-      float: right;
-
-      &:hover {
-        border-bottom: solid transparent 4px;
-      }
-
-    }
-
-    &.selected {
-      background-color: transparent;
-      border-bottom: solid $teal-dark 4px;
-    }
+  .subnav-iconBar {
+    height: 3px;
   }
 }
 
-//subnav
+.subnav-pageTitleContainer {
+  @include make-xs-column(11);
+  display: none;
+  margin-top: 5px;
+}
+
+.subnav-pageTitle {
+  color: $white;
+}
+
+.icon-bar {
+  background-color: $white;
+}
+
+.subnav-collapseContainer {
+  padding: 0 1px;
+}
+
+.subnav-item {
+  border-bottom: solid transparent 4px;
+  float: left;
+  height: 50px;
+  padding: 10px 18px;
+
+  &:hover {
+    background-color: transparent;
+    border-bottom: solid $teal-dark 4px;
+  }
+
+  &:focus {
+    background-color: transparent;
+  }
+}
+
+.subnav-item-active {
+  background-color: transparent;
+  border-bottom: solid $teal-dark 4px;
+}
+
+.subnav-link {
+  color: $white;
+  font-weight: 300;
+  padding-bottom: 16px;
+  padding-top: 12px;
+  text-decoration: none;
+
+  &:hover,
+  &:focus {
+    background-color: transparent;
+    color: $white;
+    text-decoration: none;
+  }
+}
+
+.subnav-link-active {
+  color: $white;
+  text-decoration: none;
+}
+
+.subnav-leadBtn {
+  @include button;
+  @include button-lead;
+  padding: 8px 10px;
+
+  &:focus {
+    color: $white;
+    text-decoration: none;
+  }
+}
+
+.subnav-noHover {
+  float: right;
+  padding: 8px 0;
+
+  &:hover {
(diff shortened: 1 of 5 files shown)


styleguide/_themes/derek/scss/components/subnav.scss
```diff
@@ -1,91 +1,144 @@
...
+.subnav-leadBtn {
+  @include button;
+  @include button-lead;
+  padding: 8px 10px;
+}
 
-      &:hover {
-        border-bottom: solid transparent 4px;
-      }
+.subnav-noHover > a {
```
WidgetsBurritos: This can't be applied to `.subnav-leadBtn`?

WidgetsBurritos:
Check the mobile-view. 

image

styleguide/_themes/derek/scss/components/subnav.scss
```diff
@@ -1,91 +1,148 @@
...
+    width: $full-width;
+
+    &:hover {
+      border-bottom: 0;
+      border-left: 2px solid $teal-dark;
+    }
+  }
+
+  .subnav-collapseBtn {
+    padding-right: 0px
```
WidgetsBurritos: missing semi-colon here.

WidgetsBurritos: also 0 shouldn't have units.

WidgetsBurritos:
I believe this should bump the sass lint error count down to 103.  Can you update `package.json` to reflect that?

amberRrucker:
Made those changes @WidgetsBurritos

WidgetsBurritos:
@amberRrucker What about this one? 
https://github.com/rackerlabs/zoolander/pull/216#discussion_r75879498

Also this needs a rebase.

amberRrucker:
@WidgetsBurritos yes I also made that change. That was the addition of the padding-right: 0 to the collapse button. 
I'll rebase now.

WidgetsBurritos:
No, not that.  I was referring to this: 
image

styleguide/_themes/derek/scss/components/subnav.scss
amberRrucker: No. So if I dont apply this to the li around the button then it gets the hover treatment and padding associated to all the li's. The button is the link within the li. The hover state doesnt live on those.

runrog:
👍  New active/hover state looks much better than the thick underline, verified text doesn't break funny.
subnav-mobile-refactor

styleguide/_themes/derek/scss/components/subnav.scss
WidgetsBurritos: What exactly is this supposed to be doing then, because it's supposed to be changing the focus background color?  I don't notice a difference on focus.  I've completely removed this entire selector from my code and it looks the exact same. 

Although, I do notice black text on focus, so that seems like something that should be addressed.

Normal state: 
image

Focus state:
image

amberRrucker: @WidgetsBurritos Sorry, I thought you were refering to this entire selector, not just the focus. The focus was added because we encountered an issue with the button in www and had to add it when we originally built the subnav. I will fix the color and text decoration though.

WidgetsBurritos:
👍

WidgetsBurritos:
👍  for changes.

styleguide/_themes/derek/scss/components/subnav.scss
```diff
@@ -1,91 +1,158 @@
...
-      margin-top: 10px;
-    }
+.subnav-pageTitle {
+  color: $white;
+}
 
-    button {
-      background-color: transparent;
-    }
+.icon-bar {
```
bighappyface: Should this go in `bootstrap_overrides.scss`?

styleguide/derek/_partials/solutions/subnav.ejs
```diff
@@ -1,23 +1,28 @@
...
+          <span class="sr-only">Toggle navigation</span>
+          <span class="icon-bar subnav-iconBar"></span>
+          <span class="icon-bar subnav-iconBar"></span>
+          <span class="icon-bar subnav-iconBar"></span>
+        </button>
       </div>
-      <div class="col-xs-10">
-        <!-- Needs to have the title of the page youre on -->
+      <div class="subnav-pageTitleContainer">
+        <h6 class="subnav-pageTitle">OpenStack Public Cloud</h6>
```
bighappyface: This PR is not the place to iron it out, but we should spend some time cleaning up our headings to a better document outline:

screen shot 2016-08-30 at 7 19 30 am

bighappyface:
+1

bighappyface:
Any ideas on addressing very long link labels?

screen shot 2016-08-30 at 7 29 35 am

xinbin:
"Don't? " I'm just kidding. How about an ellipsis after a set amount of
chars. Or what do others do?

styleguide/_themes/derek/scss/components/subnav.scss
WidgetsBurritos: @bighappyface The purpose of `bootstrap_overrides.scss` is to have a place for bootstrap fixes that can't be solved without excluding it from the sass linter (due to the need for weird nesting and element usage in the selector). We're trying to minimize the amount of code that gets excluded from sass lint, so we don't have a bunch of crappy code making it through, if possible.  Perhaps the file name is a bit misleading, and it should be called `no-sass-lint.scss` or something like that, but that is the point of it. 

That said, if we want somewhere to override bootstrap behavior, that is also run through the sass linter, then I'm totally fine with this idea. I would definitely agree that the `subnav.scss` file is not an appropriate place for this selector.

WidgetsBurritos:
@bighappyface I don't necessarily hate the way it does it now.  Looks a little strange, but it's not horrible (and still significantly better than what its currently doing on production). 

That said, one possible fix could be placing a max width on the `<li>` tags (plus a few other minor tweaks). This example has it set at 170px. Although, that might look weird if there are only 2-3 buttons and one is really long, and got wrapped. 

image

bighappyface:
@WidgetsBurritos I think that adjustment is preferable. A _true_ responsive design will respond to changes from the outside (viewport changes) and from the inside (content changes). It would be a good next step for our designs to consider the nature of our content, especially regarding i18n.

For instance, what if the text is instead read right-to-left? How does it handle major character shifts as we see with Chinese text?

Similar to my comments on the headings, this is probably not the PR to rethink overall strategy; however, it may be an appropriate place to make some improvements, starting with the subnav.

xinbin:
to further add to my comment. I thought it wise to check out what Spanish content site might do. I wasn't able to find a matching example but I'll attach here. Simple saying they used breadcrumbs and wrapped or maybe used a "+" 
festival___peopleenespanol_com
ponte_bella___peopleenespanol_com2
ponte_bella___peopleenespanol_com

amberRrucker:
@bighappyface I think on the subnav the best idea would be a character limit, which is unfortunate but really I think the best option given these really out of hand page titles.
but as far as rethinking our strategy around those… we’re actually going to be implementing a side navigation subnav once we go to wider than 960 on the site. The designs are completed and will support our large number of interior pages much better.

WidgetsBurritos:
@amberRrucker Character limits won't necessarily work with Smartling translations.

ejchiefs:
Given the fact that this PR is not introducing any new bugs, let's move forward and communicate this issues with Addison from a translation stand point. I will commit to pushing prioritization of moving to a fluid design that will assist in solving for problems such as these.

xinbin:
@amberRrucker hey there mademoiselle, please resolve this conflict when able.

PR #220, opened 2016-08-23
styling(forms): Added www modal forms to zoolander

runrog:
Added in modal forms from www

### DIFF
--- a/styleguide/_themes/global/scss/forms.scss
+++ b/styleguide/_themes/global/scss/forms.scss
@@ -114,3 +114,54 @@ table {
 .table > thead > tr > th {
   font-weight: unset;
 }
+
+.formModal-closeIcon {
+  font-size: 25px;
+  opacity: .7;
+  position: absolute;
+  right: 10px;
+  top: 10px;
+}
+
+.formModal-title {
+  text-align: center;
+}
+
+.formModal-content {
+  border-radius: 0;
+  padding: 15px;
+}
+
+.formModal-header {
+  border-bottom: 0;
+}
+
+.formModal-body {
+  padding: 10px 20px;
+}
+
+.formModal-label {
+  color: $gray-darkest;
+}
+
+.formModal-required {
+  color: $brand-danger;
+}
+
+.formModal-input {
+  background-color: $gray-lightest;
+  border: 1px solid transparent;
+  box-shadow: none;
+
+  &:focus {
+    border-color: $brand-secondary;
+    box-shadow: none;
+    outline: 0;
+  }
+}
+
+@media (min-width: $screen-sm-min) {
+  .formModal-dialog {
+    width: 440px;
+  }
+}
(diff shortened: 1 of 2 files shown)


styleguide/_themes/global/scss/forms.scss
```diff
@@ -114,3 +114,51 @@ table {
...
+.formModal-closeIcon {
+  opacity: .7;
+}
+
+.formModal-title {
+  text-align: center;
+}
+
+.formModal-content {
+  border-radius: 2px;
```
amberRrucker: we dont have any curved edges anywhere else. I would rather we stay consistent.

styleguide/_themes/global/scss/forms.scss
```diff
@@ -114,3 +114,51 @@ table {
...
+}
+
+.formModal-body {
+  padding: 10px 20px;
+}
+
+.formModal-dialog {
+
+  @media (min-width: $screen-sm-min) {
+    width: 440px;
```
amberRrucker: let's try not to use fixed widths.

runrog: @amberRrucker What do you recommend here? All bootstrap modals have fixed widths up until it hits tablet/mobile, at which point it goes fluid. If we go percentage width on desktop, it's either going to look too wide or too small depending on the screen size.

amberRrucker: ah. I see @runrog. That seems reasonable.

styleguide/_themes/global/scss/forms.scss
runrog: @amberRrucker oops I'll set this to 0.

runrog:
screen shot 2016-08-23 at 1 16 01 pm

amberRrucker:
@runrog Can we do an options for a 50/50 radio button section for like yes/no? And can you set up an example with a real form title? It's really likely they wont ever be that short so it may be helpful to see a longer options. I think it looks a little weird that the x is so far into the form. shouldnt it be more close to the upper right edge there?

runrog:
@amberRrucker 50/50 yes, I'll add that in. I'll add in longer text in the example title, it should just push the form elements down some. The 'x' was the OOB bootstrap close button in modals but I will adjust it, it is a little too far in for my taste also.

runrog:
screen shot 2016-08-23 at 2 14 35 pm

amberRrucker:
+1 looks good to me. @bighappyface or @WidgetsBurritos for code review?

styleguide/_themes/global/scss/forms.scss
WidgetsBurritos: Don't place the media query inside the selector.  Move it outside the selector instead.  

i.e.: 
```
  .formModal-dialog {
    width: 440px;
  }
}
```

runrog: will fix

WidgetsBurritos:
+1

PR #215, opened 2016-08-16
feat(landing-pages):first pass at landing page options

amberRrucker:
as well as video embed, form for lp, and documentation around those pages.
https://jira.rax.io/browse/RSWEB-6795

PR #221, opened 2016-08-24
RSWEB-6984: Add support for Text/Lead/Video patterns in mini-panels

WidgetsBurritos:
Jira: https://jira.rax.io/browse/RSWEB-6984

This PR is a second attempt at mini panels.  To allow for incremental change, we're only focusing on Text/Lead/Video patterns right now. 

Note: Until #215 is merged in, the video pattern will look a little strange.


amberRrucker:
Looks good to me but if there is no river color then the alignment doesnt match up because there is no padding around the non-well version. Is that something we want to care for?

WidgetsBurritos:
@amberRrucker If there's no river color, then don't have to place it in a "well".   They can just let the pattern stand on its own.

amberRrucker:
@WidgetsBurritos I mean if there is a well on the left then it gets additional padding - making the top of the content move down. On the right if there is no well then the top of the content is still further up, because of the lack of padding. So the left and right content blocks dont align in situations of mixed well and no well.

WidgetsBurritos:
@amberRrucker That's intentional. If there's a well, the content is supposed to line up with the top of the well, not the content itself.  @ejchiefs was there when we did that, but we should probably discuss with him whether or not that's the course of action we want to take.

amberRrucker:
@ejchiefs do you mind weighing in on the alignment/padding situation?
Aside from padding concerns its a +1 from me.

ejchiefs:
Yes, this is intentional. Content that is not contained within a well should not have additional padding. It should align with the top of the well. I don't think we will have instances of a lead outside of a well next to a lead inside a well since that would we have two h1s on a page. The well is meant for corner pocket content which will most likely be text/general/video until we actually create the various patterns for things like pricing.

amberRrucker:
Alrighty then I'm a +1

bighappyface:
+1

PR #225, opened 2016-08-25
RSWEB-7204: Fixes equal height on key offers

WidgetsBurritos:
This PR fixes the zoolander-side of the key offer bug.


writes:
👍

PR #226, opened 2016-08-30
fix(keyOffers): rsweb-7214 Key Offers Fix

SyntaxStacks:


WidgetsBurritos:
image

+1.  I can confirm this fixes it.

amberRrucker:
Those icons shouldnt be that dark should they? Theyre normally lighter I think.

xinbin:
@amberRrucker , aye it does seem like the hover gray is darker now. Not sure when that change snuck in.

PR #228, opened 2016-08-31
RSWEB-7224: update padding for value props

writes:
Updates padding for value props.


amberRrucker:
The icon and hover color have both changed at some point and we need to fix that. Hover color should be gray-lighter and icon should be gray-darker.

amberRrucker:
Do you mind adding some text length variation to one of these so we can make sure the equal height rows are working?

writes:
@amberRrucker I added in one long one. Let me know if you see anything else. Thanks for the feedback!

WidgetsBurritos:
👍

amberRrucker:
Thank you. Looks good. +1
@WidgetsBurritos

PR #232, opened 2016-09-02
RSWEB-7241: equal height fix to resource class

writes:
Updates resource class to ```display: flex;``` so that background will extend full height of the row.

### DIFF
--- a/styleguide/_themes/derek/scss/patterns/grid/businessResource.scss
+++ b/styleguide/_themes/derek/scss/patterns/grid/businessResource.scss
@@ -49,3 +49,7 @@
   @include make-sm-column(6);
   @include flex-box;
 }
+
+.resource {
+  @include flex-box;
+}


WidgetsBurritos:
👍

amberRrucker:
In the PR description in the future can you include the page where its viewable - for posterity sake?

amberRrucker:
+1 - looks good to me. Does anybody know when it broke?

WidgetsBurritos:
@amberRrucker It came when we refactored these components a week or so ago. I wasn't able to use all the maintainable classes because how the grid patterns are so tightly coupled, so until they get decoupled we're gonna have some issues.

writes:
Yep will do. I chose to put it in the ticket instead for this one since this is a public repo and the page is not yet live with an alias. I will remember for future pages that are currently live to add it in.

PR #233, opened 2016-09-02
RSWEB-7205: focus state color should not change in footer links

writes:
Changes the focus state to be the original link color and removes the underline. Also removes the change to black for the active state in the footer.

Can be seen here http://localhost:9000/derek/examples/events

### DIFF
--- a/styleguide/_themes/derek/scss/globalElements/footer.scss
+++ b/styleguide/_themes/derek/scss/globalElements/footer.scss
@@ -18,9 +18,13 @@
   font-size: 1.2rem;
   text-decoration: none;
 
+  &:focus {
+    text-decoration: none;
+  }
+
+  &:focus,
   &:hover {
     color: $gray-light;
-    text-decoration: underline;
   }
 }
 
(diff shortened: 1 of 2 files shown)


WidgetsBurritos:
Drupal is still adding a li a.active that is getting precedence over this...

image

styleguide/_themes/derek/scss/globalElements/footer.scss
```diff
@@ -18,9 +18,13 @@
   font-size: 1.2rem;
   text-decoration: none;
 
+  &:focus {
+    text-decoration: none;
+  }
+
+  &:focus,
   &:hover {
     color: $gray-light;
```
WidgetsBurritos: unfortunately drupal is still setting `a:focus` for these.  This will need to get updated in bootstrap overrides as well. 
image

writes:
@WidgetsBurritos This is ready for another review. Thanks for the feedback!

runrog:
👍  as far as coding style

WidgetsBurritos:
👍   

[Before:]
screen shot 2016-09-09 at 3 40 46 pm

[After:]
screen shot 2016-09-09 at 3 41 05 pm

PR #236, opened 2016-09-06
RSWEB-7250: update subnav padding for larger viewports

writes:
Updates padding for the subnav when in larger viewports (>1200px width).

View on this page locally. http://localhost:9000/derek/examples/office365


WidgetsBurritos:
👍 

Zoolander (Before): 
image

Zoolander (After):
image

www (Before): 
image

www (After): 
image

PR #239, opened 2016-09-13
bug(nav): reduced padding for translations in nav

runrog:
Reduced padding on nav items so the nav doesn't break
screen shot 2016-09-13 at 10 56 54 am


SyntaxStacks:
👍

PR #240, opened 2016-09-14
RSWEB-7235: update mobile banner min height

writes:
Updates min banner height for mobile to avoid long banner text being cut off or the banner background showing below the subnav.

- example page: https://www.rackspace.com/managed-aws

- Corresponding www PR with margin top being removed. (1276)


WidgetsBurritos:
👍

PR #248, opened 2016-09-21
RSWEB-6905: zoolander documentation

amberRrucker:
RSWEB-7353

Creating patterns documentation for our available patterns.
Viewable here: http://localhost:9000/derek/examples/solutions-usage
and linked to from here: http://localhost:9000/derek/examples/solutions

Looking for clear and concise documentation but not many changes to actual patterns other than adjusting dummy text to be within character limits.

### DIFF
--- a/styleguide/derek/examples/layouts/additional-resources.ejs
+++ b/styleguide/derek/examples/layouts/additional-resources.ejs
@@ -1,20 +1,18 @@
 <!-- Additional Resources: 33/34/33 -->
 <div class="container half-padding-full">
-  <h4>Additional Resources</h4>
   <div class="row row-eq-height">
     <div class="col-sm-4 panel-flex">
       <%- partial('../../_partials/cards/additional-resource.ejs') %>
     </div>
     <div class="col-sm-4 panel-flex">
       <%- partial('../../_partials/cards/additional-resource.ejs', {
-        type: 'video',
-        desc: 'This is short'
+        type: 'video'
       }) %>
     </div>
     <div class="col-sm-4 panel-flex">
       <%- partial('../../_partials/cards/additional-resource.ejs', {
         type: 'rss',
-        desc: 'This is extra long. This is extra long. This is extra long. This is extra long. This is extra long. This is extra long. This is extra long. This is extra long. This is extra long. This is extra long.'
+        desc: 'consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean m'
       }) %>
     </div>
   </div>
@@ -24,8 +22,7 @@
     </div>
     <div class="col-sm-6 panel-flex">
       <%- partial('../../_partials/cards/additional-resource.ejs', {
-        type: 'support',
-        desc: 'This is short'
+        type: 'support'
       }) %>
     </div>
   </div>
--- a/styleguide/derek/examples/layouts/announcements.ejs
+++ b/styleguide/derek/examples/layouts/announcements.ejs
@@ -1,27 +1,27 @@
 <!-- Announcements -->
 <div class="container">
-  <h4>Announcements</h4>
   <div class="row row-eq-height">
     <div class="col-sm-12">
       <%- partial('../../_partials/cards/announcements.ejs', {
         'type': 'newProduct',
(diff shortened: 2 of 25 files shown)


styleguide/derek/examples/solutions.ejs
```diff
@@ -2,22 +2,7 @@
...
-  </div>
-  <div class="col-md-6">
-    <h4 class="red">DONT</h4>
-    <ul class="times">
-      <li>bold, italicize or alter any text in any way other than centering</li>
-      <li>include buttons within any patterns other than the cta</li>
-      <li>put in any hard coded styles on any elements within patterns</li>
-    </ul>
-  </div>
+  <p>For more information on how exactly to use our guidelines please see <a href="/derek/solutions-usage">Solutions Usage</a></p>
```
amberRrucker: We want everyone to be able to easily view all of our patterns that are available without having to scroll through all the usage information, especially with the ever growing list of patterns. As we add them it would be nice to have a page to see them all as they would on a page and then a page that's more documentation heavy.

styleguide/derek/examples/layouts/announcements.ejs
```diff
@@ -1,24 +1,22 @@
 <!-- Announcements -->
 <div class="container">
-  <h4>Announcements</h4>
   <div class="row row-eq-height">
     <div class="col-sm-12">
       <%- partial('../../_partials/cards/announcements.ejs', {
-        'type': 'newProduct',
-        'desc': 'This is a New Product. Vivamus quis sem sollicitudin, cursus tellus eget, tempor ex. Suspendisse ut magna condimentum, blandit sem vel, auctor urna. Aenean eu auctor erat. '
```
amberRrucker: It's available options are listed right above it on the usage page, but for available solutions that makes sense as well. I would rather add it as the title that we have in the backlog to add to the pattern though, I'll note to transition that when that work gets done. For now I'll add back in that line.

styleguide/derek/examples/layouts/additional-resources.ejs
```diff
@@ -1,20 +1,17 @@
...
       <%- partial('../../_partials/cards/additional-resource.ejs', {
-        type: 'video',
-        desc: 'This is short'
+        type: 'video'
       }) %>
     </div>
     <div class="col-sm-4 panel-flex">
       <%- partial('../../_partials/cards/additional-resource.ejs', {
-        type: 'rss',
-        desc: 'This is extra long. This is extra long. This is extra long. This is extra long. This is extra long. This is extra long. This is extra long. This is extra long. This is extra long. This is extra long.'
```
amberRrucker: Im adding shortened lengths but not super long ones because the ipsum ive put in is the max allowable characters so I dont want to show anything longer than that.

styleguide/derek/examples/solutions.ejs
WidgetsBurritos: If that's a concern I'd prefer having some sort of anchor system into the particular patterns instead of having the patterns spread across multiple pages. We could potentially adapt what I did on the subpanels page for this. Or maybe have an option to collapse usage information. To me it seems ambiguous and confusing having it spread across multiple pages. 

But I think we should get feedback from the rest of the team on this.

bighappyface: After looking at the current docs/content organization in Zoolander I favor clarifying what is there before adding new pages. I am not sure this perspective adds much to this conversation, but I can say that when I was looking at things yesterday the examples are already behind and what we have isn't that great.

amberRrucker: What do you mean by the examples are behind and what we have isn't great? Are there updates to the patterns that happened outside of zoolander?

styleguide/derek/examples/layouts/additional-resources.ejs
WidgetsBurritos: That's fine.

styleguide/derek/examples/layouts/announcements.ejs
WidgetsBurritos: Cool

runrog:
👍 on design. I would love it if we can talk about organizing this in the near future

amberRrucker:
@runrog that's the plan stan

PR #253, opened 2016-09-27
fix(banners): rsweb-7409 IMAC banners mobile BG fix

SyntaxStacks:


WidgetsBurritos:
+1 - need to confirm it in staging. 

image

image

image

PR #256, opened 2016-09-28
RSWEB-7397: imac-banner-button-padding-only

amberRrucker:
RSWEB-7397

Reducing button padding for imac banner button ONLY to 8px top/bottom
Visible here:
http://localhost:9000/derek/examples/imac


WidgetsBurritos:
This technically already exists, so we could theoretically handle it separately, but changing this padding seems to accentuate how off the caret looks on the button. Are you seeing this on your end? 

Before:
image

After:
image

runrog:
👍 Design, buttons working/looking as intended 
screen shot 2016-09-28 at 1 27 50 pm

SyntaxStacks:
QE 👍

PR #259, opened 2016-09-28
rsweb-7420-table-fontsize

amberRrucker:
bug(zoolander):table font size overridden by www

Can be viewed here (but should be the same):
http://localhost:9000/derek/examples/tables

Fixes issues here: 
https://admin.rackspace.com/node/5725


WidgetsBurritos:
👍  - needs design review @runrog

runrog:
👍 Desgin
screen shot 2016-09-29 at 9 41 47 am

PR #260, opened 2016-09-29
RSWEB-7368: center img on ias pattern

writes:
Updates IAS pattern on the solutions page to be consistent with the counterpart on www. 

**Will not require a www counterpart and has no affect on the live site.** 

http://localhost:9000/derek/examples/solutions


WidgetsBurritos:
👍

SyntaxStacks:
QE +1

PR #263, opened 2016-09-29
rsweb-7441-subhead-switch

amberRrucker:
RSWEB-7441
style(zoolander):subhead font switch

Can be viewed on homepage and solutions examples page to check for any weirdness.

### DIFF
--- a/styleguide/_layouts/default.jade
+++ b/styleguide/_layouts/default.jade
@@ -16,7 +16,7 @@ html(lang="en")
 
     link(href="#{ css }", rel="stylesheet")
     link(href="css/main.css", rel="stylesheet")
-    link(href="https://fonts.googleapis.com/css?family=Open+Sans:700,300,200,600,800,400|Khand:700,600,500|Montserrat:400,700", rel="stylesheet")
+    link(href="https://fonts.googleapis.com/css?family=Open+Sans:700,300,200,600,800,400", rel="stylesheet")
     link(href="https://maxcdn.bootstrapcdn.com/font-awesome/4.6.1/css/font-awesome.min.css" rel="stylesheet")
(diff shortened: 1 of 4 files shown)


styleguide/_layouts/default.jade
```diff
@@ -16,7 +16,7 @@ html(lang="en")
 
     link(href="#{ css }", rel="stylesheet")
     link(href="css/main.css", rel="stylesheet")
-    link(href="https://fonts.googleapis.com/css?family=Open+Sans:700,300,200,600,800,400|Khand:700,600,500|Montserrat:400,700", rel="stylesheet")
+    link(href="https://fonts.googleapis.com/css?family=Open+Sans:700,300,200,600,800,400|Khand:700,600,500", rel="stylesheet")
```
ejchiefs: Can remove Khand here since we are pulling it locally now. This will probably be in conflict.

ejchiefs:
👍

styleguide/_themes/global/scss/fonts.scss
```diff
@@ -13,10 +13,11 @@ $code: Tahoma, Arial, Geneva, Helvetica;
 @mixin subhead($color: $gray-darkest) {
   color: $color;
   font-family: $subhead;
-  font-weight: 400;
+  font-weight: 500;
+  letter-spacing: .35px;
```
WidgetsBurritos: I'm not sure px are the appropriate unit here. Probably `em` would be best so it's always relative to the font size.

styleguide/_themes/global/scss/fonts.scss
```diff
@@ -13,10 +13,11 @@ $code: Tahoma, Arial, Geneva, Helvetica;
...
   line-height: 1.45em;
   text-align: left;
-  text-transform: uppercase;
+  text-transform: capitalize;
```
ejchiefs: We should just remove text-transform here and use exactly what gets inserted into the CMS. There might be instances where ECMs want to use a product name that starts with lowercase.

PR #265, opened 2016-09-30
RSWEB-7448: Add style to the stories

WidgetsBurritos:
This PR adds html/css for the customer stories pages into zoolander.


styleguide/derek/examples/index.jade
```diff
@@ -69,4 +69,13 @@
...
     .col-md-4
+      h5 Story Overview
+        a(href="derek/examples/stories")
+          img.img-thumbnail(src="imgs/derek/examples/story-overview.png")
+  .row
+    .col-md-4
+      h5 Story Interior
+        a(href="derek/examples/stories-interior")
+          img.img-thumbnail(src="imgs/derek/examples/story-interior.png")
+    .col-md-4
```
amberRrucker: In order to slim this stuff down, can we assume if they get to the overview page of events or stories that they can get to the interior and remove those from the examples page?

styleguide/derek/examples/stories-interior.ejs
```diff
@@ -0,0 +1,49 @@
...
+<%- partial("../_partials/_mainnav") %>
+<%- partial('../_partials/solutions/header-smallIMAC', { headline: 'Customer Stories'}) %>
+
+<!-- Story Region -->
+<div class="container standard-padding">
+  <div class="row">
+    <div class="story-colLeft">
+      <h1 class="story-title">Under Armour</h1>
+      <p>The company is the developer, marketer, and supplier of a wide range of sportswear and casual apparel mainly focusing on hi-tech sportswear for professional athletes. Under Armour began offering footwear in 2006, and now operates as a global company with offices in Denver, Amsterdam, Denver, Hong Kong, Toronto and Guangzhou, China.</p>
+      <p><a href="javascript: void(0);" class="story-caseStudy-button">Read the Case Study</a></p>
```
amberRrucker: I thought we were avoiding a third part of maintainable classes. So it would be story-caseStudyButton, unless the third part was a state.

styleguide/derek/examples/stories-interior.ejs
```diff
@@ -0,0 +1,49 @@
+<%- partial("../_partials/_ceiling") %>
+<%- partial("../_partials/_mainnav") %>
+<%- partial('../_partials/solutions/header-smallIMAC', { headline: 'Customer Stories'}) %>
```
amberRrucker: Did we care for this on the imac example page for banners?

styleguide/_themes/derek/scss/snowflakes/stories.scss
```diff
@@ -0,0 +1,94 @@
...
+
+.story-title {
+  text-align: left;
+}
+
+.story-caseStudy-button {
+  @include button;
+  @include button-learning;
+  margin-bottom: 1rem;
+  width: auto;
```
amberRrucker: why auto? Shouldnt it work like the rest of the site until we make that change globally? Also is there a reason we want the button to have a bottom margin instead of the container just have some bottom padding?

styleguide/_themes/derek/scss/snowflakes/stories.scss
```diff
@@ -0,0 +1,94 @@
...
+  padding: 0 .6rem;
+}
+
+.storyOverview-itemInner {
+  border: 1px solid $gray-light;
+  padding: 5px;
+}
+
+.storyOverview-logo {
+  background-color: $gray;
```
amberRrucker: shouldnt this be gray-base?

amberRrucker:
screen shot 2016-09-30 at 1 31 30 pm
I'd prefer the title like this instead if possible?

styleguide/derek/examples/index.jade
WidgetsBurritos: Good call. I'll adjust.

styleguide/derek/examples/stories-interior.ejs
WidgetsBurritos: Wait, what?

styleguide/derek/examples/stories-interior.ejs
WidgetsBurritos: You're right.  I'll adjust.

styleguide/derek/examples/stories-interior.ejs
amberRrucker: http://localhost:9000/derek/examples/imac
http://localhost:9000/derek/assets/photography

Those pages have small banners that were broken by this update to the partial.

styleguide/_themes/derek/scss/snowflakes/stories.scss
WidgetsBurritos: This was an exact copy of the buttons we used on events. 
The reason for the `width: auto;` was this: 

screen shot 2016-09-30 at 1 43 17 pm

vs 

screen shot 2016-09-30 at 1 43 30 pm

styleguide/_themes/derek/scss/snowflakes/stories.scss
WidgetsBurritos: yep. good catch.

styleguide/_themes/derek/scss/snowflakes/stories.scss
amberRrucker: Yeah. so either we need to set it into a container or we do this until we change the widths on buttons entirely. Since this would be the only two places we're changing the way we set widths Im concerned with the impending change to buttons and ctas that this may end up wonky. @ejchiefs you want to weight in on this one?

styleguide/derek/examples/stories-interior.ejs
WidgetsBurritos: oh weird. That's because it's jade calling an ejs file.  Hmm.  Have you dealt with that before?  Not sure if there's an easy solution for that. I could just pull the imac banner code directly into the template instead.

JesseCastro:
👍  Code looks clean, downloaded and built locally, zero errors observed.  Nice commit, looks very good.  Props @amberRrucker - great design/interaction on this new style page!

styleguide/_themes/derek/scss/snowflakes/stories.scss
amberRrucker: Sounds like this got talked about. Im good here.

amberRrucker:
If we can figure out the problem with the imac examples page then im good. +1

styleguide/derek/examples/stories-interior.ejs
WidgetsBurritos: @amberRrucker,  The new github review system doesn't seem to hide messages once they've been addressed any more.  I just pulled the HTML directly into the template.  

https://github.com/rackerlabs/zoolander/pull/265/files#diff-a07e96d6486c531c335851c2ea60f322R13

amberRrucker: perfect! +1

PR #272, opened 2016-10-05
RSWEB-7467: [bug] Adjust size of search button

WidgetsBurritos:
This PR adjusts search button size to fix issues with spanish.

### DIFF
--- a/styleguide/_themes/derek/scss/globalElements/mainnav.scss
+++ b/styleguide/_themes/derek/scss/globalElements/mainnav.scss
@@ -259,7 +259,7 @@
   border: 1px solid $gray-lighter;
   outline: none;
   padding: 3px;
-  width: 80%;
+  width: 72%;
 }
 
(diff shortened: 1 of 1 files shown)


WidgetsBurritos:
Before:
screen shot 2016-10-05 at 10 52 01 am

After:
screen shot 2016-10-05 at 10 51 53 am

runrog:
👍  design
Looks god in mobile as well
screen shot 2016-10-05 at 11 00 53 am

WidgetsBurritos:
@SyntaxStacks, maybe but that would require refactoring as there are other issues in play here, which would need a separate ticket.  The bugfix task itself should simply fix the bug in question.

SyntaxStacks:
I believe a task for size adjustment is the perfect time to refactor code that adjust to the size of the content. I'll defer to EJ on this as this is technically good to go.

PR #273, opened 2016-10-06
RSWEB-7499: [bug] Adjusts font size to fit inside smaller container

WidgetsBurritos:
This PR adjusts font size for the rsweb icon on features.


ejchiefs:
👍

PR #275, opened 2016-10-10
RSWEB-7323: Kill page wrap

WidgetsBurritos:
This PR kills the `margin-top` on the imac and events banners.  This is tied to killing the page wrap and www and must go out with that PR.


bighappyface:
+1

SyntaxStacks:
👍

PR #278, opened 2016-10-12
RSWEB-7528: homepage adjustments

amberRrucker:
style(homepage):adjusting for 1280px move

RSWEB-7528

Changes can be seen on this page:
http://localhost:9000/derek/examples/homepage

Looking for styling changes to make sure nothing breaks/looks terrible in our move to 1280 instead of 960

### DIFF
--- a/styleguide/_themes/derek/scss/snowflakes/homepage.scss
+++ b/styleguide/_themes/derek/scss/snowflakes/homepage.scss
@@ -114,16 +114,16 @@
 }
 
 .hostingOptions-logo {
-  @include make-sm-column(5);
-  @include make-sm-column-push(3.5);
+  @include make-sm-column(3.5);
+  @include make-sm-column-push(4.25);
   @include make-xs-column(3);
   @include make-xs-column-push(4.5);
   text-align: center;
 }
 
 .hostingOptions-logoWide {
-  @include make-sm-column(10);
-  @include make-sm-column-push(1);
+  @include make-sm-column(7);
+  @include make-sm-column-push(2.5);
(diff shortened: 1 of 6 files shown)


styleguide/_themes/derek/scss/snowflakes/homepage.scss
```diff
@@ -92,16 +92,16 @@
 }
 
 .hostingOptions-logo {
-  @include make-sm-column(5);
-  @include make-sm-column-push(3.5);
+  @include make-sm-column(3.5);
```
WidgetsBurritos: I understand the need for push's to have fractional offsets, when dealing with an odd number of columns, but do we really want columns themselves to not adhere to sizing standards? That doesn't seem like a good idea to me, and could start being treated as precedent, which I don't think we want.

amberRrucker: I feel like the homepage is sort of a special case, since it needs to be so graphically wonky to work. Also, soon this section will go away and it wont be an issue anymore. Unfortunately right now, I don't really see another way to accomplish this?

WidgetsBurritos: ehh....  I guess it's fine for now, since we're not looking to refactor everything.  But wanted to clarify that I don't necessarily like it and we shouldn't try hard to avoid doing this moving forward.

xinbin:
Please rebase. <3  so I can rebuild this in a test branch

SyntaxStacks:
Rebased this work @rackerlabs/rackspace-web-team

PR #280, opened 2016-10-13
RSWEB-7529: pattern(announcement): added header and hover state if link

runrog:
Needs design/code review. Added a header option and wrapped linked announcements in a custom a tag.
screen shot 2016-10-13 at 1 49 52 pm

### DIFF
--- a/styleguide/_themes/derek/scss/patterns/announcements.scss
+++ b/styleguide/_themes/derek/scss/patterns/announcements.scss
@@ -18,9 +18,33 @@
   }
 }
 
+.announcement-link {
+  line-height: inherit;
+  text-decoration: none;
+
+  &:hover,
+  &:focus {
+    color: $blue-gray;
+    text-decoration: none;
+  }
+}
+
+.announcement-linkText {
+  display: block;
+  font-weight: bold;
+  margin-top: 10px;
+}
+
+.announcement-link-hover {
+
+  &:hover {
+    background-color: $gray-lightest;
+    color: $blue-gray;
+  }
+}
+
 .announcement-message {
   @include make-md-column(10);
-  @include flex-box;
   padding: 12px 15px;
 }
 
@@ -48,5 +72,13 @@
   .announcement-icon,
   .announcement-message {
     width: $full-width;
+  }
+
+  .announcement-head {
+    text-align: center;
+  }
+
+  .announcement-linkText {
+    text-align: center;
   }
 }
--- a/styleguide/derek/_partials/cards/announcements-link.ejs
+++ b/styleguide/derek/_partials/cards/announcements-link.ejs
@@ -9,10 +9,13 @@ var icons = {
   'promotion' : 'fa-money'
 };
 %>
-<div class="announcement-wrapper announcement-<%- type %> fade in row-eq-height">
-  <div class="announcement-icon flex-center-all"><i class="fa <%- icons[type] %>" aria-hidden="true"></i></div>
-  <div class="announcement-message">
-    <h3 class="announcement-head">Header Here</h3>
-    <%- desc %>
+<a href="javascript: void(0)" class="announcement-link">
+  <div class="announcement-wrapper announcement-<%- type %> fade in row-eq-height announcement-link-hover">
+    <div class="announcement-icon flex-center-all"><i class="fa <%- icons[type] %>" aria-hidden="true"></i></div>
+    <div class="announcement-message">
+      <h3 class="announcement-head">Header Here</h3>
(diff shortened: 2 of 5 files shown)


styleguide/_themes/derek/scss/patterns/announcements.scss
```diff
@@ -49,4 +57,8 @@
   .announcement-message {
     width: $full-width;
   }
+
+  .announcement-head {
+    text-align: center;
```
WidgetsBurritos: Any thing we do on the mobile version, needs to be added to the subpanel version as well.  I realize announcements aren't currently working on zoolander inside the subpanels page, but basically all that needs to happen is your mobile code should also be placed here:

https://github.com/rackerlabs/zoolander/blob/master/styleguide/_themes/derek/scss/patterns/panels.scss#L14

runrog:
@amberRrucker design review por favor

amberRrucker:
I was thinking more that the box would get an overlay or something when hovered over if there is a link. We also may want to still have space for the link text since mobile wont have a hover. Run it by mike with the text highlight and see what he thinks.

runrog:
@amberRrucker I initially wanted to do a bg hover but it gets tricky since an announcment pattern that is a link will be getting wrapped in an 'a' tag and non links won't. a dynamic hover class would probly need to be added on the pattern container if it's a link. I'm going to push up another commit soon with an example (as requested by Stinemetze)

WidgetsBurritos:
+1 on 19d014cb946604b0c99a1f6ac7d5cebe4df72f99

amberRrucker:
Let's not have the paragraph change color on hover. It's a description, not a link so let's just give the whole area the gray hover state and make sure there is only a hover state for ones that have a link. So add a text link to one of them as the linked example.

WidgetsBurritos:
👍  on 4179b2ed0e0bfc1d1e449e9eeb502f191f98af48

amberRrucker:
The text is still highlighting on hover in the standard announcement examples.
Don't worry about showing them in the splits as those examples get shown on the subpanels page.
Please add link text to the linked version. Mobile won't have the hover so we need another visual cue that it's linking.

WidgetsBurritos:
Also, you may want to do something about the focus state.  That looks funky:

image

styleguide/_themes/derek/scss/patterns/announcements.scss
```diff
@@ -18,9 +18,25 @@
   }
 }
 
+.announcement-link {
+  line-height: inherit;
+  text-decoration: none;
+
+  &:hover {
+    text-decoration: none;
```
WidgetsBurritos: add `&:focus` here

styleguide/derek/_partials/cards/announcements-link.ejs
```diff
@@ -0,0 +1,21 @@
...
+  'promotion' : 'fa-money'
+};
+%>
+<a href="#" class="announcement-link">
+  <div class="announcement-wrapper announcement-<%- type %> fade in row-eq-height announcement-link-hover">
+    <div class="announcement-icon flex-center-all"><i class="fa <%- icons[type] %>" aria-hidden="true"></i></div>
+    <div class="announcement-message">
+      <h3 class="announcement-head">Header Here</h3>
+      <%- desc %>
+      <span href="#" class="announcement-linkText">View More</span>
```
WidgetsBurritos: No `href` here.

styleguide/derek/_partials/cards/announcements-link.ejs
```diff
@@ -0,0 +1,21 @@
...
+if(!desc) { var desc = "Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec quam felis, ultricies nec, pellentesque eu, pretium."; }
+if(!type) { var type = 'general'; }
+var icons = {
+  'alert' : 'fa-exclamation-circle',
+  'general' : 'fa-bullhorn',
+  'newProduct' : 'fa-server',
+  'promotion' : 'fa-money'
+};
+%>
+<a href="#" class="announcement-link">
```
WidgetsBurritos: can you set the href to be `javascript: void(0)` instead?

runrog:
@WidgetsBurritos done sir

amberRrucker:
+1

WidgetsBurritos:
Adding `HOLD` until www counterpart is complete.

PR #287, opened 2016-10-25
rsweb-7627-reduce-open-sans

amberRrucker:
perf(fonts):reducing open sans usage
RSWEB-7627

Reducing the number of weights we're using and adding in variables for weights to try and reduce the number of weights that get added in the future.


WidgetsBurritos:
```
styleguide/_themes/derek/scss/components/ticker.scss:  font-weight: 700;
styleguide/_themes/derek/scss/snowflakes/homepage.scss:    font-weight: 300;
styleguide/_themes/global/scss/banners.scss:  font-weight: 700;
styleguide/_themes/global/scss/bootstrap_variables.scss:$headings-font-weight:    500 !default;
styleguide/_themes/global/scss/bootstrap_variables.scss:$btn-font-weight:                400 !default;
styleguide/_themes/global/scss/bootstrap_variables.scss:$close-font-weight:           bold !default;
styleguide/_themes/global/scss/fonts.scss:  font-weight: normal;
styleguide/_themes/global/scss/rsweb-font-style.scss:    font-weight: normal;
styleguide/_themes/global/scss/rsweb-font-style.scss:    font-weight: normal;
```

Should we adjust any of these?

amberRrucker:
I left ticker and legacy code for homepage and the old banner since those code chunks need to get removed anyways. I'll adjust the bootstrap variables but do we need to change the font-weight: normal for icons?

WidgetsBurritos:
@amberRrucker I dunno.  I just did a quick search for font weights still explicitly referenced. Maybe it would make sense to have a $normal variable for instances like this.  I think the goal is to get rid of the font-weight variations.

amberRrucker:
@WidgetsBurritos added a normal and switched the naming up a bit to not have multiple $black variables.

PR #296, opened 2016-11-04
RSWEB-7579: remove uneeded margin mobile

writes:
Remove unneeded margin on features pattern in mobile.

http://local.rackspace.com:7080/openstack/private/enablement


WidgetsBurritos:
👍  on the condition that we put in a separate ticket to figure out the breakpoint issues.

PR #305, opened 2016-11-14
RSWEB-7825: [1280] Change breakpoint from 1200 to 1280

WidgetsBurritos:
We need to set are largest breakpoint from 1200 to 1280 if that's actually are max width. This prevents side scrolling issues related to the breakpoint.  

Additionally since `.rug-container` sets the left and right margins to zero, we can't use the `.row` class in here.  I'm creating a www PR for this fix as well. 

And also, the basement padding should be set to match the gutter width.

### DIFF
--- a/styleguide/_themes/derek/scss/globalElements/basement.scss
+++ b/styleguide/_themes/derek/scss/globalElements/basement.scss
@@ -16,7 +16,7 @@
 
 // Basement Classes
 .basement {
-  padding: 10px 15px;
+  padding: 10px $grid-gutter-width;
 }
 
 .basement-copyright {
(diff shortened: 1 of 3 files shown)


styleguide/_themes/derek/scss/globalElements/basement.scss
```diff
@@ -16,7 +16,7 @@
 
 // Basement Classes
 .basement {
-  padding: 10px 15px;
+  padding: 10px $grid-gutter-width;
```
SyntaxStacks: `$grid-gutter-width` is 30px. Is this what we want, as the previous value was 15px.
https://github.com/rackerlabs/zoolander/blob/59e620b11f6807d622b69f17d5c8f8a88ddf4d25/styleguide/_themes/global/scss/bootstrap_variables.scss#L352

PR #306, opened 2016-11-15
RSWEB-7824: Remove max height

WidgetsBurritos:
This PR removes the max height from the customer logos


WidgetsBurritos:
Before:
screen shot 2016-11-15 at 2 35 09 pm

After:
screen shot 2016-11-15 at 2 35 02 pm

runrog:
👍 on design

PR #312, opened 2016-11-17
style(rsweb-7848-title-case): remove uppercase from headings

ejchiefs:
This PR converts headings to title case by removing the text-transform attribute from the headings mixin and adding it to the imacBanner-heading to ensure we retained uppercase for our banners.


WidgetsBurritos:
👍

brantdarilek:
👍

PR #313, opened 2016-11-18
RSWEB-7868: Make region selector great again

WidgetsBurritos:
This PR fixes the region selector on mobile by pushing the text into multiple lines instead of trying to stack them in a small space.

### DIFF
--- a/styleguide/_themes/derek/scss/globalElements/basement.scss
+++ b/styleguide/_themes/derek/scss/globalElements/basement.scss
@@ -26,7 +26,7 @@
 
 .basement-copyright {
   color: $gray-base;
-  text-align: right;
+  text-align: center;
 }
 
 .basementLp-copyright {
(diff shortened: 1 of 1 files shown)


styleguide/_themes/derek/scss/globalElements/basement.scss
```diff
@@ -26,7 +26,7 @@
 
 .basement-copyright {
   color: $gray-base;
-  text-align: right;
+  text-align: center;
```
brantdarilek: Moving copyright to the center of the page?

WidgetsBurritos: just on mobile.  See the media query below.

amberRrucker:
Very clean solution. +1

brantdarilek:
👍

PR #315, opened 2016-11-18
RSWEB-7082: remove id selectors

brantdarilek:


brantdarilek:
I agree, but I think its going to depend on how they are being used. I'm going to look at content on www and determine if thats necessary or not.

brantdarilek:
@WidgetsBurritos 
Adrianna pull pages that had these ids and all of those are now using other styles to overwrite these. This should be good to go.

WidgetsBurritos:
+1 

image

 That's a pretty sight. Almost there.

PR #316, opened 2016-11-18
RSWEB-7083: sass lint class errors

WidgetsBurritos:
This PR cleans up class name sass-lint issues.

Note: The customer-stories.ejs is old (there's another file with the updated style), so I just removed the unnecessary code. 

And since there is js attached to `.chat_hidden` I just moved it to an excluded location.


brantdarilek:
👍

bighappyface:
+1

PR #322, opened 2016-11-22
RSWEB-7085: Sass-lint nesting issues

WidgetsBurritos:
This PR addresses `warning  Nesting depth 3 greater than max of 2  nesting-depth` sass lint errors on the following files.  Here is how I did it. A few assumptions were made so this will need to be verified separately. 

- homepage.scss: I just removed all of the old style code as its no longer used. @amberRrucker is this correct? 
- o365.scss: By forcing `td` to be a direct child of `tr`, it removes a level of nesting from sass-lint's perspective.
- fonts.scss: By forcing `li` to be a direct child of `.green` or `.teal` lists, it removes a level of nesting from sass-lint's perspective.

An important thing to note is that I left the nesting issues on the `tabs.scss` file. The reason for this is that the tabs are getting refactored next couple of sprints, so we can just let that process handle the cleanup of those errors instead as it is likely to be all brand new code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "test:visual:init": "nohup npm start & (cd ./node_modules/backstopjs && npm run reference)",
     "backstop-update": "node ./scripts/update-backstop-urls.js",
     "sass:compile": "node-sass --output-style=compressed --quiet -o /tmp styleguide/css",
-    "sass-lint": "sass-lint -v --max-warnings 19",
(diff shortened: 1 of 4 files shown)


writes:
+1

WidgetsBurritos:
Amber gave me a verbal +1.

SyntaxStacks:
👍

PR #326, opened 2016-12-01
refactor(region-selector): replace flags and logo for performance

brantdarilek:
Not sure if I'm placing the flag image in the right place. But the image itself is "global" and the other styles are unique to Derek. So let me know if this works.


styleguide/_themes/global/scss/shared_assets.scss
```diff
@@ -0,0 +1,5 @@
...
+}
```
WidgetsBurritos: I definitely like having a place to share these types of resources in the code so we don't define them multiple times.

WidgetsBurritos:
👍 works great for me.

bighappyface:
+1 :shipit:

PR #327, opened 2016-12-02
RSWEB:7692 - feat(zoolander): new tabs

runrog:
Created completely new tabs, we will also need a PR in www for the js and will need to migrate our old tabs to these, then remove the old ones from zoolander as well.

Screenshots from Zoolander's solutions page. Added a bottom red border that wasn't in the original design per Tim's recommendation as it would be easier to see where the tabs ended when there is content underneath them.

screen shot 2016-12-02 at 11 17 36 am
screen shot 2016-12-02 at 11 17 53 am


styleguide/_themes/derek/scss/components/rsTabs.scss
```diff
@@ -0,0 +1,126 @@
...
+    position: relative;
+    z-index: 5;
+
+    &:hover {
+      background-color: $rackspace-red;
+    }
+
+    .rsTabs-tabLink {
+      color: $white;
+    }
```
WidgetsBurritos: Is all this nesting necessary?

styleguide/_themes/derek/scss/components/rsTabs.scss
```diff
@@ -0,0 +1,126 @@
...
+  padding: 10px 0;
+
+  .rsTabs-tabActive {
+    display: block;
+  }
+}
+
+[class^='rsTabs-tabContent-'] {
+  display: none;
+}
```
WidgetsBurritos: This going to be a performance issue.  I'd recommend having two classes on the tabs: 

`rsTabs-tabContent` and then `rsTabs-tabContent-1`. 
`rsTabs-tabContent` would have the `display: none` set and then just let JS handle the display of it.

styleguide/_themes/derek/scss/components/rsTabs.scss
```diff
@@ -0,0 +1,126 @@
...
+  .rsTabs-tabMobile {
+    display: block;
+  }
+
+  .rsTabs-contentContainer {
+    border: 1px solid $gray-light;
+    border-bottom: 2px solid $rackspace-red;
+    border-top: 0;
+    padding: 5px;
+  }
```
WidgetsBurritos: Nesting necessary here too?

styleguide/_themes/derek/scss/components/rsTabs.scss
```diff
@@ -0,0 +1,126 @@
...
+.rsTabs-contentContainer {
+  border-bottom: 2px solid;
+  border-color: $rackspace-red;
+  border-top: 2px solid;
+  min-height: 10px;
+  padding: 10px 0;
+
+  .rsTabs-tabActive {
+    display: block;
+  }
```
WidgetsBurritos: Nesting necessary here too?

styleguide/js/zoolander.js
```diff
@@ -139,6 +141,70 @@ jQuery(document).ready(function($) {
...
+
+    $tabs.each(function(){
+      var $thisTabSet = $(this);
+      var $tabLink = $thisTabSet.find('.rsTabs-menu .rsTabs-tabItem');
+
+      //function to reset the tabs on page resizes
+      var resizeTimer;
+      $zoolanderWindow.resize(function(){
+        clearTimeout(resizeTimer);
+        resizeTimer = setTimeout(resizeResetTabs, 250);
```
WidgetsBurritos: I do like the idea of a timeout here to prevent if from re-firing a bunch.  I'm just wondering about the time frame.  Would it make more sense to set a lower threshhold, like `100` to minimize the delay? More of a question than a suggestion.

styleguide/js/zoolander.js
```diff
@@ -139,6 +141,70 @@ jQuery(document).ready(function($) {
...
+          }
+        }
+      });
+      $tabLink.find('a').click(function(e){
+        e.preventDefault();
+      });
+    });
+    return this;
+  };
+}(jQuery));
```
WidgetsBurritos: I'm know we don't currently have it built into zoolander, but I'd really like to see some sort of testing around this functionality. Maybe we can add some end to end testing on this separately.

styleguide/js/zoolander.js
```diff
@@ -139,6 +141,70 @@ jQuery(document).ready(function($) {
...
+        $tabLink.each(function(){
+          var $tL = $(this);
+          if($tL.hasClass('rsTabs-tabActive')){
+            $tL.siblings().removeClass('rsTabs-tabMobile');
+          }
+        });
+      }
+
+      //set first tab active
+      $thisTabSet.find('.rsTabs-menu .rsTabs-tabItem').first().addClass('rsTabs-tabActive');
```
WidgetsBurritos: If tab items are always going to be direct descendants of the menu does this make more sense for a performance impact? 
```
$thisTabSet.find('.rsTabs-menu > .rsTabs-tabItem')
```

Or maybe just search for the destination itself:
```
$thisTabSet.find('.rsTabs-tabItem')
```

This is a question more than a suggestion. Just trying to create a discussion around that.

styleguide/_themes/derek/scss/components/rsTabs.scss
runrog: Same here, .rsTabs-tabActive is used in the menu and the content container, but they need to behave a little differently so I nested them under the different areas.

styleguide/_themes/derek/scss/components/rsTabs.scss
runrog: 👍  good point will get that done

styleguide/_themes/derek/scss/components/rsTabs.scss
runrog: yes this is the mobile version of the content container.. it's nested in a media query

styleguide/js/zoolander.js
runrog: I can switch this to 100ms also. Based on my experience it's not visually a noticeable difference though. I've tried this technique with 10 before and it still looks the same. Would be nice if we could see a diff in delay

styleguide/js/zoolander.js
runrog: Thats a good question. I believe 
```
$thisTabSet.find('.rsTabs-menu > .rsTabs-tabItem')
```
would be the quicker option, once jquery finds .rsTabs-menu it then searches for .rsTabs-tabItem inside the menu only.

This option
```
$thisTabSet.find('.rsTabs-tabItem')
```
Will search the entire tabs for .rsTabs-tabItem, even inside the content area where they aren't there.

styleguide/js/zoolander.js
runrog: 👍  on this

runrog:
@WidgetsBurritos As for the existing bootstrap input. Not really because this new design isn't going to be using an accordion interaction on mobile and the bootstrap js out of the box doesn't have any mobile solutions for tabs. We would need to right new js on top of it. So rather than fight with it, I created new ones.

WidgetsBurritos:
image

I don't really know if it is obvious enough that there are other tabs here. I mean I guess there's a down arrow icon, but it confused me at first when I saw it.  

Should it not just display like this and place the content block underneath the selected tab instead? 
image

If we decide to keep it the way it is, then that's fine, but just figured I'd voice my opinion on that.  

If we do decide to keep it that way, the down arrow itself doesn't appear to be clickable (just the red region around it)
image

runrog:
@WidgetsBurritos thanks for the input. I hear what you're saying, this was considered and went through design/ux review. But if the tabs are open like that, it will create a longer scroll experience and it may cause someone to hit a button while scrolling too.

styleguide/_themes/derek/scss/components/rsTabs.scss
bighappyface: Does `.rsTabs-tabActive` need to be nested within `.rsTabs-menu` though?

runrog: Since .rsTabs-tabActive does something different inside the menu than it does in the content section I nested it inside both. I'll just add another class

WidgetsBurritos:
@runrog Okay if that's what they want, then that's fine.  We do still need to make the arrow clickable though.

runrog:
@WidgetsBurritos refactored to make the mobile icon clickable

styleguide/js/zoolander.js
```diff
@@ -139,6 +141,73 @@ jQuery(document).ready(function($) {
...
+          var $tL = $(this);
+          if($tL.hasClass('rsTabs-tabActive-li')){
+            $tL.siblings('li').removeClass('rsTabs-tabMobile');
+          }
+        });
+      }
+
+      //set first tab active
+      $tabLink.first().addClass('rsTabs-tabActive-li');
+      $thisTabSet.find('.rsTabs-contentContainer div:first-child').addClass('rsTabs-tabActive');
```
WidgetsBurritos: This needs `>`

runrog: done

WidgetsBurritos:
👍 on this.  I do think we probably need to iterate on this in the future from a JS perspective, but I think this works for now.

ejchiefs:
@runrog / @amberRrucker, can we discuss the mobile implementation? There are some potential UX concerns that I am seeing in the interactive versions that weren't clear in the mockup. The easiest solution in my opinion is to replicate the default behavior of an accordion menu on mobile. By doing this, we would have all tabs shown at all times with the active pane visible underneath its parent tab rather than hiding the tabs behind an arrow icon and placing the active pane underneath all tabs. Happy to discuss offline to clarify.

runrog:
@ejchiefs @amberRrucker lets talk about it in the Deliotte standup if we have time

amberRrucker:
@ejchiefs I agree with you. I think the default behavior of an accordion would be excellent. I hadnt yet taken a look at this one.

styleguide/_themes/derek/scss/components/tabs.scss
```diff
@@ -1,94 +1,111 @@
...
-  text-align: center;
+}
 
-  & > li {
-    background-color: $teal-dark;
-    border-top: 5px solid $teal-dark;
-    display: inline-block;
-    float: none;
-    padding: 0;
+.nav-tabs-li {
```
WidgetsBurritos: can this be name spaced under `.rsTabs`?

styleguide/_themes/derek/scss/components/tabs.scss
```diff
@@ -1,94 +1,111 @@
...
-      border-top: 5px solid $white;
+  &:first-child {
+    margin-left: 10px;
+  }
 
-      a {
-        background-color: $teal-dark;
-      }
-    }
+  & > .nav-tabs-a {
```
WidgetsBurritos: `.rsTabs-tabLink`?

Also, does this have to be nested here, or can it just be placed in its own top-level item?   I'm assuming its because of some weird bootstrap `li > a` declarations, but just trying to make sure.

styleguide/derek/solutions/solutions.ejs
```diff
@@ -41,10 +41,18 @@
...
 
+<!-- Rs Tabs -->
+<div class="container">
+  <h4>Rs tabs</h4>
+  <div class="row">
+    <div class="col-xs-12">
+      <%- partial('../_partials/solutions/tabs.ejs') %>
+    </div>
+  </div>
+</div>
```
WidgetsBurritos: Can you throw all this wrapping code into a `../layouts/tabs.ejs` file like we've done with some of the other patterns? This will help us keep the solutions page relatively clutter free.

styleguide/derek/_partials/solutions/tabs.ejs
```diff
@@ -1,47 +1,14 @@
...
+  <div id="myTabContent" class="tab-content rsTabs-content">
+      <div class="tab-pane active" id="home">
+          <p>Raw denim you probably haven't...</p>
       </div>
-    </div>
-    <div role="tabpanel" class="tab-pane" id="tab-6">
-      <div class="container">
-        Content for tab 6
+      <div class="tab-pane" id="profile">
+          <p>Food truck fixie locavore, accus...</p>
```
WidgetsBurritos: Mmmmmm.... slider provider.

styleguide/_themes/derek/scss/components/tabs.scss
```diff
@@ -1,94 +1,111 @@
...
+    border-bottom: 0;
+    border-color: transparent;
+    border-radius: 0;
```
WidgetsBurritos: Can you use `@border-radius(0);` so we can pull in the vendor prefix support?

styleguide/_themes/derek/scss/components/tabs.scss
```diff
@@ -1,94 +1,111 @@
...
+}
 
-      &:hover,
-      &:active {
-        border-top: 5px solid $gray-darkest;
+// Overrides for the tabCollapse plugin & bootstrap panel classes
+.rsTabs-container {
+  .panel {
+    border-left: 1px solid $gray-lighter;
+    border-radius: 0;
```
WidgetsBurritos: `@include border-radius(0);`

runrog:
Latest look of the tabs - Utilizing the old tabCollapse plugin. 

screen shot 2016-12-08 at 2 39 43 pm
screen shot 2016-12-08 at 2 40 15 pm

styleguide/_themes/derek/scss/components/tabs.scss
```diff
@@ -1,94 +1,114 @@
...
 
-    &.active {
-      background-color: $white;
+  &.active {
+    background-color: $rackspace-red;
+    border: 0;
+    color: $white;
+    top: 0;
+
+    & .rsTab-link {
```
WidgetsBurritos: Should this have a `>`?

styleguide/derek/_partials/solutions/tabs.ejs
```diff
@@ -1,47 +1,308 @@
...
-      <div class="container">
-        Content for tab 5
-      </div>
-    </div>
-    <div role="tabpanel" class="tab-pane" id="tab-6">
-      <div class="container">
-        Content for tab 6
-      </div>
-    </div>
+<div class="rsTabs-container">
```
WidgetsBurritos: Will you place some sort of separator, or maybe a comment explaining that this is showing what it looks like with an excessive amount of tabs? It's a little confusing to see where one ends and the other starts.

WidgetsBurritos:
So apparently part of the issue I described above is due to bootstrap having precedent:

image

WidgetsBurritos:
Oh actually the issue appears to be on desktop. Even though the failover is supposed to be for tablet, I think we should do it regardless of whether or not we're on tablet or desktop. Otherwise we get issues like above.

package.json
```diff
@@ -17,7 +17,7 @@
     "test:visual:init": "nohup npm start & (cd ./node_modules/backstopjs && npm run reference)",
     "backstop-update": "node ./scripts/update-backstop-urls.js",
     "sass:compile": "node-sass --output-style=compressed --quiet -o /tmp styleguide/css",
-    "sass-lint": "sass-lint -v --max-warnings 8",
+    "sass-lint": "sass-lint -v",
```
WidgetsBurritos: 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈 🎈

runrog: 🎆  🎊  🎊 🔔 ❗️ ❗️ 💯  😱

ejchiefs:
👍

runrog:
👍 on @WidgetsBurritos new commit

PR #329, opened 2016-12-02
feat(rsweb): use sass map for rsweb icon classes

bighappyface:
Any value in this?


WidgetsBurritos:
Seems easier to read in my opinion.  Are there any other reasons for doing this or just general cleanup?

brantdarilek:
Easier to read, IMO.

bighappyface:
I saw a snippet months back about handling character fonts using a map and thought to add it here for fun.

writes:
Like it

WidgetsBurritos:
+1

PR #332, opened 2016-12-07
RSWEB-7921: min height for row to be at least two lines

writes:
Sets min height for Additional Resources pattern to be at least 2 lines tall. 

http://localhost:9000/derek/solutions/solutions

`.additional-resource .row-eq-height { min-height: 78px; }`

### DIFF
--- a/styleguide/_themes/derek/scss/patterns/cards/resources.scss
+++ b/styleguide/_themes/derek/scss/patterns/cards/resources.scss
@@ -107,6 +107,7 @@
   .row-eq-height {
     flex-grow: 2;
     margin: 10px;
+    min-height: 7.8rem;
   }
 }
 


amberRrucker:
Instead of pixels do we want to use rem so that when we alter the text size it will adjust to be the proper height of two lines?

WidgetsBurritos:
@amberRrucker is it `rem` or just `em`?  Are we planning on changing font sizes globally or just for the pattern itself?

amberRrucker:
@WidgetsBurritos rem will change it based on the root font size, which is what we want since we're making the font adjustment globally, and in the future we probably want it based off of that. Do you agree?

WidgetsBurritos:
@amberRrucker Yup.  I just wasn't sure if we were changing font sizes globally or just in the pattern. If it's a global change then `rem` is definitely the correct choice there. @writes You mind adjusting?

writes:
@amberRrucker @WidgetsBurritos I have updated the `px` to `rem`. Thanks!

amberRrucker:
+1

PR #333, opened 2016-12-07
RSWEB-7996: Remove partner/developer links

WidgetsBurritos:
This PR removes the developer and partner links from the ceiling. This work technically only needs to be done on www, but I did it here as well for consistency.


ejchiefs:
👍

amberRrucker:
@WidgetsBurritos +1
Although we dont really have to do it because im removing them in my nav work right now anyways.

PR #334, opened 2016-12-08
RSWEB-7811: Remove row and adjust width for features

WidgetsBurritos:
This PR fixes a few layout issues with features:

1.) Adds `width: 100%;` to the pattern wrapper.
2.) Remove the `<div class="row row-eq-height">` from inside the pattern. This has to be done on the `www` side as well. 
3.) Removed the margin-bottom from the features pattern, as all margins/patterns between panes is now handled by the layouts instead of the pattern itself.
4.) Added `subpanel` and `subpanel-inner` classes to get zoolander markup to match up with www.
5.) Added the features pattern to the solutions page as it was missing. 

Note: Because this is just a bug fix, I stopped short of actually refactoring features, which I believe is in queue to be tackled in the not-too-distant future. For that reason, I did the minimal I deemed necessary to make zoolander useful for the features pattern. This is a www counterpart PR for this task to handle the above markup change.

### DIFF
--- a/styleguide/_themes/derek/scss/patterns/cards/productFeatures.scss
+++ b/styleguide/_themes/derek/scss/patterns/cards/productFeatures.scss
@@ -1,6 +1,6 @@
 .features-wrapper {
   @include flex-box;
-  margin-bottom: 1em;
+  width: 100%;
 
   &.features-threeUp {
     display: block;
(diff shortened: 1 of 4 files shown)


WidgetsBurritos:
Before: 
screen shot 2016-12-07 at 10 17 41 pm

After: 
screen shot 2016-12-07 at 10 17 49 pm

runrog:
👍

PR #335, opened 2016-12-08
RSWEB-7802: Remove 300 font weight, and increase line-height/base font height.

amberRrucker:
style(fonts):removing 300 font weight, increasing line height and base font height.

Can be previewed across the styleguide. Should check here:
http://localhost:9000/derek/solutions/solutions 
and here:
http://localhost:9000/derek/assets/typography

### DIFF
--- a/styleguide/_layouts/default.jade
+++ b/styleguide/_layouts/default.jade
@@ -16,7 +16,7 @@ html(lang="en")
 
     link(href="#{ css }", rel="stylesheet")
     link(href="css/main.css", rel="stylesheet")
-    link(href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600", rel="stylesheet")
+    link(href="https://fonts.googleapis.com/css?family=Open+Sans:400,600", rel="stylesheet")
     link(href="https://maxcdn.bootstrapcdn.com/font-awesome/4.6.1/css/font-awesome.min.css" rel="stylesheet")
     script(src="https://code.jquery.com/jquery-1.12.3.min.js" integrity="sha256-aaODHAgvwQW1bFOGXMeX+pC4PZIPsvn2h1sArYOhgXQ=" crossorigin="anonymous")
     script(src="https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/zoolander/modernizr.min.js")
--- a/styleguide/_themes/derek/scss/patterns/cards/keyOffers.scss
+++ b/styleguide/_themes/derek/scss/patterns/cards/keyOffers.scss
@@ -24,7 +24,6 @@
 }
 
 .keyoffersWidget-body {
-  line-height: 1.7em;
   padding-bottom: 10px;
 }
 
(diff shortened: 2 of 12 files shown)


styleguide/_themes/derek/scss/patterns/cards/keyOffers.scss
```diff
@@ -24,7 +24,7 @@
 }
 
 .keyoffersWidget-body {
-  line-height: 1.7em;
+  //line-height: 1.7em;
```
WidgetsBurritos: Should we just remove this line instead?

styleguide/_layouts/default.jade
```diff
@@ -16,7 +16,7 @@ html(lang="en")
 
     link(href="#{ css }", rel="stylesheet")
     link(href="css/main.css", rel="stylesheet")
-    link(href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600", rel="stylesheet")
+    link(href="https://fonts.googleapis.com/css?family=Open+Sans:400,600", rel="stylesheet")
```
WidgetsBurritos: Do you know, is there a www ticket to remove the font weight there as well?

amberRrucker:
@WidgetsBurritos you're super right and I just removed that line.

WidgetsBurritos:
👍 from a code perspective.

WidgetsBurritos:
Running through this with EJ we found a few issues:

The rug is a bit awkward now:
screen shot 2016-12-08 at 2 35 06 pm

The nav has a weird bottom border:
screen shot 2016-12-08 at 2 35 23 pm

Also it appears lead content (underneath the title) are still getting a font-weight: 300 from somewhere. Not entirely sure where it's coming in, but we're still seeing it in there.

amberRrucker:
I've fixed the issues with the navigation and the rug. There is a quick and dirty fix for the rug that I'm not sure will fly with everyone but the alternatives are .... annoying when we're taking out the rug soon anyways?
@WidgetsBurritos @ejchiefs

bighappyface:
I am still getting the bottom padding in the main navigation.

screen shot 2016-12-09 at 10 06 17 am

amberRrucker:
@bighappyface, 
I don't think I was looking at it quite wide enough. Should be good now.

styleguide/_themes/global/scss/bootstrap_overrides.scss
```diff
@@ -47,6 +47,15 @@
   }
 }
 
+//Quick and Dirty fix before we take out the rug.
+.fa-3x {
+  font-size: 2.5em!important;
```
WidgetsBurritos: If we're gonna do something hacky, I'd rather do something like this instead:
```
.rug-container .fa-3x { font-size: 2.5em; }
```

WidgetsBurritos: If we're gonna do something hacky, I'd rather do something like this instead:

```
.rug-container .fa-3x { font-size: 2.5em; }
```

amberRrucker:
@WidgetsBurritos got it! thanks.

bighappyface:
Desktop looks good.

The nav is still a bit off for mobile:

screen shot 2016-12-09 at 12 45 48 pm
screen shot 2016-12-09 at 12 45 30 pm

amberRrucker:
@bighappyface alrighty! got it. Thank you.

runrog:
👍 much smoother

PR #337, opened 2016-12-12
RSWEB-8009 [www] Infinity Numbers not displayed when the website is viewed on mobile

bighappyface:
This PR sets the `font-size` to "0" for `.navbar-phoneicon` to hide inner text used by Infinity call tracking "autodiscovery". The phone number text has to be put in the element given class `.navbar-phoneicon` for autodiscovery to work but on mobile only the icon should be visible.


WidgetsBurritos:
👍

brantdarilek:
👍

PR #339, opened 2016-12-14
RSWEB-8061: Fix line-height and font related issues

WidgetsBurritos:
This PR fixes some issues found with the latest font weight and line height changes. 

This requires a www ticket as well.


ejchiefs:
👍

PR #340, opened 2016-12-15
Rsweb 7734 sexy nav

amberRrucker:
feat(sexy-nav):new navigation


amberRrucker:
Need to go through the js to see what we need and don't around the search functionality. I wasnt sure what would be helpful or in the way with our current implementation. Also need to do some squashing. Figure we can chat about this together @WidgetsBurritos

styleguide/_layouts/default.jade
```diff
@@ -82,3 +82,6 @@ html(lang="en")
   script(src="https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/zoolander/bootstrap-tabcollapse.js")
   script.
     $(".imacBanner-headline").fitText(1.1, { maxFontSize: '45px' });
+  script(src="js/classie.js")
```
WidgetsBurritos: we talked about this separately, but just including for the sake of the review. Since we're already somewhat coupled to jQuery, let's just rely on that for now. If we decide to remove jQuery this may be a direction to go with in the future, but I'd rather not add any additional libraries to the code if we have an alternative already built in.

styleguide/derek/_partials/_mainnav.ejs
```diff
@@ -1,14 +1,14 @@
...
-          </div>
-          <a class="navbar-searchicon navbar-searchDesktop hidden-xs" id="searchDesktop"><span class="navbar-activeArrow"></span></a>
+          <a class="navbar-hamburger collapsed" data-toggle="collapse" data-target="#main-navigation"></a>
+          <div id="navbar-search" class="navbar-search hidden-xs">
+        		<form>
+        			<input class="navbar-search-input" placeholder="Search" type="text" value="" name="search" id="search">
+        			<input class="navbar-search-submit" type="submit" value="">
+        			<span class="navbar-icon-search"></span>
+        		</form>
+        	</div>
```
WidgetsBurritos: Can we fix some of the whitespace here?

styleguide/js/zoolander.js
```diff
@@ -189,3 +176,128 @@ var url = window.location.pathname;
...
+		close : function() {
+			this.inputEl.blur();
+			classie.remove( this.el, 'navbar-search-open' );
+		}
+	}
+
+	// add to global namespace
+	window.UISearch = UISearch;
+
+} )( window );
```
WidgetsBurritos: I'll wait to review this JS until I see your updates.

WidgetsBurritos:
On slightly older phones (iPhone 5), the menu does this: 

image

If we get down to that small, should we maybe kill the text on the buttons and just show the phone/email/chat icons?

WidgetsBurritos:
I'm still seeing the nav peaking over the edge of the content on the right hand side in mobile.

image

amberRrucker:
hey @WidgetsBurritos what browser are you seeing that on? I ran it through browserstack and I'm not seeing that issue. 
https://www.browserstack.com/screenshots/f97c34d5e2e73410983f5eb11051368ff2e02155

WidgetsBurritos:
I see it in Chrome when in compatibility mode. I also ran it on my phone via ngrok and got the same issue

WidgetsBurritos:
@amberRrucker It seems to be an issue on any page with a footer and basement.  So you're not gonna see it on the homepage for example, but it does happen on the stories/events page. It may be a bug in those other patterns, but given that production has these components on it, it will probably need to be fixed as part of this effort.

styleguide/js/zoolander.js
```diff
@@ -189,3 +180,17 @@ var url = window.location.pathname;
...
+var $searchSubmit = $searchContainer.find('.navbar-icon-search');
+
+$searchSubmit.click(function() {
+  if ($searchContainer.hasClass('expanded')) {
+    $searchContainer.css({'width':'0'}).removeClass('expanded');
+  }
+  else {
+    $searchContainer.css({'width':'18%'}).addClass('expanded');
+  }
+});
```
WidgetsBurritos: I'd think all of this functionality should live where the old functionality lived (i.e. instead the dom ready), and not just sitting here separately.

styleguide/_themes/derek/scss/globalElements/mainnav.scss
```diff
@@ -484,3 +382,91 @@
...
+  float: right;
+  list-style: none;
+  margin-bottom: 0;
+  padding-left: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+}
+
+.menuInline-itemMobile {
```
bighappyface: Did these lose the `navbar` prefix?

styleguide/_themes/derek/scss/globalElements/mainnav.scss
```diff
@@ -275,23 +198,14 @@
...
-  }
-
-  .navbar-activeArrow {
-    margin: 0 0 0 -5px;
-  }
-}
-
 //main nav on desktop
-@media only screen and (min-width: $screen-md-min) {
+@media only screen and (min-width: $screen-xs-max) {
```
WidgetsBurritos: Should this be `$screen-xs-max`, or `$screen-sm-min` since it's a `min-width` situation?  I know we've had some weird edge-case situations, but I don't think this is one of them, but I will test these scenarios more thoroughly later .

styleguide/_themes/global/scss/bootstrap_overrides.scss
```diff
@@ -80,10 +79,33 @@
...
+
+.navbar-search-open > .navbar-icon-search,
+.no-js > .navbar-search > .navbar-icon-search {
+  background: $white;
+  color: $gray-darker;
+  z-index: 11;
+}
+
+.navbar-search-open .navbar-search-submit,
+.no-js > .navbar-search > .navbar-search-submit {
```
WidgetsBurritos: The bootstrap overrides are getting a bit ridiculous.  Not your fault I realize, but we need to discuss removing the offending code from bootstrap altogether.

WidgetsBurritos:
image

On tablet mode, should the text be left-aligned with the logo?

bighappyface:
+1 to move to www implementation

WidgetsBurritos:
👍  for now.  We may need to make a few tweaks as we implement on www, but we can address those concerns as we see them.

brantdarilek:
I'm getting a side scroll issue on the iPhone 6 and up.
screen shot 2017-01-19 at 9 50 42 am

brantdarilek:
I know we have a few of css validation errors. But if we could clean up those related to the nav.
screen shot 2017-01-19 at 9 59 45 am

amberRrucker:
What's going on with the logo @brantdarilek did something change with that recently?

brantdarilek:
@amberRrucker I just base64 encoded the previous SVG to resolve issues with IE. (some character in that SVG that IE doesn't like).

WidgetsBurritos:
@amberRrucker @brantdarilek The new logo was a PNG, and I base 64 encoded that.  If it's looking weird now, we may need to adjust a css value or something.

amberRrucker:
@WidgetsBurritos its getting smashed at lower sizes. Can we use an svg?

WidgetsBurritos:
Yes.  If you have a SVG for that, then just run it through a base 64 encoder and then we can update it.

WidgetsBurritos:
👍  on e2673c962b55b3220fbb1cedbd6aa2bc54ba19b9

WidgetsBurritos:
👍  on bbb5c493c1411e7113229b5ea2868f42642cc210

PR #341, opened 2016-12-15
RSWEB-8078: update homepage to use quadrant.jpg instead of quadrant.png

brantdarilek:
I've optimized both the png and jpg on the cdn but this should help a bunch.

screen shot 2016-12-15 at 12 32 26 pm


ejchiefs:
👍

WidgetsBurritos:
👍

PR #343, opened 2017-01-03
RSWEB-8024: refactor valueProps pattern

writes:
[zoo]
https://jira.rax.io/browse/RSWEB-8024 - https://github.com/rackerlabs/zoolander/pull/343

http://local.rackspace.com:7080/managed-hosting

[www]
https://jira.rax.io/browse/RSWEB-8029 - https://github.rackspace.com/rswebteam/www/pull/1533


WidgetsBurritos:
👍 but definitely needs www counterpart. If we get www in sync with this, the alignment issues should get fixed.

WidgetsBurritos:
👍

brantdarilek:
👍

styleguide/derek/layouts/value-props.ejs
```diff
@@ -10,7 +10,7 @@
       }) %>
     </div>
```
amberRrucker: Should we change the name of these files to reflect the valueProp name change?

amberRrucker:
Value Prop pattern should actually not contain a link. The EBPs are aware and there should be no value props on the site utilizing a link or button. Those should be all i&S if they have a link.

screen shot 2017-01-10 at 10 31 56 am
screen shot 2017-01-10 at 10 32 12 am

styleguide/derek/layouts/value-props.ejs
WidgetsBurritos: Probably not be a bad idea.

writes:
Hey @amberRrucker thank you! I have made the updates requested. Let me know if you see anything else.

amberRrucker:
+1 thanks JT.

PR #346, opened 2017-01-04
dependency issue

xinbin:
Up for discussion. A dependency of `harp` is `terraform`, which in turn has a dependency of `node-sass`
An error is thrown complaining that the version for node-sass is not found.

Submitting this for discussion. Do we need to update the dependencies?

Coming from a fresh install, I could not run npm install with out getting the following errors:
Cannot download "https://github.com/sass/node-sass/releases/download/v3.4.2/darwin-x64-51_binding.node":
HTTP error 404 Not Found

---- we could go with the solution Paul suggested , which is to run `npm install -g harp@next`


WidgetsBurritos:
Closing till we get this in a sprint

PR #348, opened 2017-01-06
RSWEB-7901: Fixed Height IMAC Banners

amberRrucker:
RSWEB-7901
Adjusting the IMAC banners to have a fixed height. There are a couple concerns around length of title ect that is being addressed with content changes vs code changes.
Visible here:
http://localhost:9000/derek/incubation/imac
and any sandbox pages


WidgetsBurritos:
screen shot 2017-01-06 at 1 17 25 pm

"BECAUSE ONE SIZE DOESN’T FIT ALL" indeed...

WidgetsBurritos:
Also, since we're trying to get a bit more uniform and expected behavior out of this, what are your thoughts about just killing fittext as well?

WidgetsBurritos:
More weirdness in mobile.  


image

amberRrucker:
@WidgetsBurritos fittext still serves the purpose of scaling for different screen sizes and retaining the line breaks in the places they should be. I'd rather keep it and reduce risk for a relatively simple task. Also you're seeing the problems in that screen shot because you need khand installed on your system. There is a huge difference between arial and khand thats causing visual differences.

WidgetsBurritos:
Not necessarily needing to be tackled now, but more for discussion. Should we set zoolander to load in khand like www does then? Because enforcing font dependencies on an individual system doesn't necessary make for friendly universal reviews.

WidgetsBurritos:
I installed Khand.  The CTA is still falling off. 

image

amberRrucker:
@WidgetsBurritos I'm going to get with Mike to see if we're utilizing the button at all in banners at this point. Remove it if we arent, adjust if we are.

WidgetsBurritos:
👍 although now that there's a change to HTML markup, we're gonna need a `www` PR to go with this.

amberRrucker:
@WidgetsBurritos theres a follow up ticket in for Roger already

runrog:
@amberRrucker @WidgetsBurritos reviewing now. On the CTA in banner discussion. We are currently running a test on the homepage with a CTA in it. Since chats declined so much when it was removed we are seeing if there are any improvements. So we may in face be keeping them if so.

amberRrucker:
@runrog yeah we already talked about all that last week. That's why we made two different options. One without the CTA and one with the CTA for homepage, why rack and landing pages.

runrog:
@amber What about translation issues? I'm going to predict that some headers will be longer in other languages and cut off.

screen shot 2017-01-23 at 10 37 15 am
screen shot 2017-01-23 at 10 37 22 am

amberRrucker:
@runrog feel free to run some of our headlines (homepage and why rack) through google translate to get a real estimate because I highly doubt it's going to add 5 lines ya crazy. We do have a character limit in english so try to use real headline options or stay within those character limits for examples please.

runrog:
@amber if you say so. other than that, it looks good in all viewports. 👍

PR #349, opened 2017-01-10
RSWEB-8033: refactor cta pattern

writes:


WidgetsBurritos:
👍

amberRrucker:
Something is up. It should look like the second one and it def does not. Not totally sure what happened there.

screen shot 2017-01-11 at 10 48 42 am
screen shot 2017-01-11 at 10 48 50 am

writes:
@WidgetsBurritos @amberRrucker I have updated the pattern. Let me know if you see any issues. 

Thanks!

styleguide/derek/_partials/solutions/cta.ejs
```diff
@@ -1,21 +1,26 @@
...
+      </div>
+    </div>
+    <div class="cta-row">
+      <div class="cta-buttonWrapper">
+        <a class="cta-button">Talk to a Specialist</a>
+      </div>
+    </div>
+    <div class="cta-row">
+      <div class="cta-subheadWrapper">
+        <h3>Call <a href="javascript:void(0)">1-855-715-7307</a> to start saving</h3>
```
bighappyface: Why can't this be a `p`?

writes: Hey @bighappyface the update has been made.

styleguide/_themes/global/scss/cta.scss
```diff
@@ -0,0 +1,40 @@
...
+}
+
+.cta-row {
+  @include make-row;
+}
+
+.cta-subheadWrapper {
+  @include make-md-column(12);
+
+  p {
```
bighappyface: Being that there seem to be specific styles for this text, I think it is appropriate to have a specific class for it. Such a class can be in the CTA module, but something tells me that this combination of family, size and weight may be useful elsewhere. If not, it's a snowflake.

writes:
@WidgetsBurritos and @bighappyface would it make sense to do `div.subheadWrapper` as opposed to `.subheadWrapper p` to account for any tags put in the field or would we rather restrict to `p` and `a` with styles set on `p`? The `p` tag default styles will currently override the class alone and the subhead uses a different font-family.

styleguide/_themes/global/scss/cta.scss
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.cta-row {
+  @include make-row;
+}
+
+.cta-subheadWrapper {
+  @include make-md-column(12);
+
+  .cta-subhead {
```
bighappyface: Is this nested for greater specificity to get the CSS properties to take effect?

amberRrucker:
Looks good. +1

styleguide/_themes/global/scss/cta.scss
writes: Yes sir. It is needed to overwrite the default `p` styles.

bighappyface:
+1

PR #350, opened 2017-01-17
RSWEB-8205: base64 encode nav logo

brantdarilek:
This PR base64 encodes the rackspace logo in the navigation. There is an offending character that is causing the logo not to render in IE11 only. This fixes that issue.


WidgetsBurritos:
👍

bighappyface:
+1

PR #351, opened 2017-01-23
feat(npm): add files to package.json

bighappyface:
This PR defines the files that should be preserved in within `node_modules/zoolander` when Zoolander is installed with NPM. The `dist` directory is ignored by default so installing it via NPM skips that directory.


WidgetsBurritos:
👍

PR #352, opened 2017-01-27
RSWEB-8026: refactor lead classses

writes:
Refactors lead pattern.

- removes `.lead` class and uses `.lead-wrapper` so that there is no need to overwrite bootstrap class 
- structure is in alignment with CTA refactor and uses maintainable
- does not alter subnav elements that still use original `.lead` class

http://localhost:9000/derek/solutions/subpanels


runrog:
👍

PR #354, opened 2017-01-27
RSWEB-8031: refactor quote pattern

writes:


writes:
This pattern is going to be redone using feeds and this refactor will likely not be needed. Cancelling for now.

PR #356, opened 2017-01-31
RSWEB-8405: Move CSS from general content type into repo

WidgetsBurritos:
This PR pulls in a few minor style updates that are currently set inside `<style></style>` tags in a general content type. 

@amberRrucker can you please confirm this is the correct solution to this problem, instead of just removing the on page style altogether?


brantdarilek:
👍

amberRrucker:
+1

PR #359, opened 2017-02-06
RSWEB-8096: Fixes gutter related issues.

WidgetsBurritos:
This PR fixes a few gutter related issues.  

1. d54c7941ae27c7cc1fd40764fedfe145637868e5 - Gets rid of containers around subnav and specifies a better gutter width for subnav items
2. 3a383650582df5bbd81de168ca0ab3908c92bf37 - Adjusts padding on half-padding/standard-padding when viewing in 1280px+
3. 068b4854d45d1bd6d22bec68044fd7f2796ce46d - Sets a margin-right of `$grid-gutter-width` unless you're in 1280px + gutter width*2. 

Can be tested in zoolander here: http://localhost:9000/derek/incubation/office365
One thing to keep in mind (and part of the reason for item 3 above, is to give room for the chat slider button.)

### DIFF
--- a/styleguide/_themes/derek/scss/components/subnav.scss
+++ b/styleguide/_themes/derek/scss/components/subnav.scss
@@ -4,7 +4,8 @@
 }
 
 .subnav-container {
-  @include make-container;
+  margin: 0 auto;
+  max-width: $screen-md-max;
   padding: 0;
 }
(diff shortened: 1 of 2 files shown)


runrog:
👍 design

PR #361, opened 2017-02-07
RSWEB-8385: Remove unused npm packages and references

WidgetsBurritos:
Not sure if there's anything else that can be scrapped, but here's a good start.


brantdarilek:
👍

PR #364, opened 2017-02-14
RSWEB-8403: consolidate font awesome and webfonts

brantdarilek:


brantdarilek:
After discussion, the team has decided to go with https://github.com/rackerlabs/zoolander/pull/365 as the preferred approach.

PR #366, opened 2017-02-15
RSWEB-8476: change checkmark color

brantdarilek:


amberRrucker:
@brantdarilek Is the goal to change all the checks on the website to green? It was a little difficult to follow on the ticket but within features and things I'm not seeing any green checks. Only gray. http://localhost:9000/derek/solutions/solutions-usage

brantdarilek:
@amberRrucker Its my understanding that these are not used and/or will be getting removed for bullets. I've also discussed and confirmed with Mike.

amberRrucker:
+1

PR #370, opened 2017-02-21
rsweb-8520-update-quote-pattern

amberRrucker:
RSWEB-8520
Updating quote pattern to be more readable and separating fields in the attribution. Did some refactoring while working on this. 

Viewable for the quote and quote split here:

Design is in task RSWEB-8408


WidgetsBurritos:
image

Should logo be centered on mobile?

styleguide/_themes/derek/scss/patterns/quote.scss
```diff
@@ -1,3 +1,68 @@
+.quote {
+  background-color: $blue-gray;
+  padding: $half-padding;
+}
+
+.quote-container {
+  @include make-container;
```
WidgetsBurritos: Are we ever placing these quotes in mini panels, because if so we shouldn't be defining a container here, and should probably just rely on standard bootstrap containers for that like we do with the other patterns.

amberRrucker:
@WidgetsBurritos the padding on the left should be the same as the footer links. Theres not actually any right hand padding. I'm not sure on the mini panel stuff. I just kept the container with the padding and background, because the color should be standard across all quotes and it was already on all of them so I didn't want there to be more things to change in the CMS.

WidgetsBurritos:
@amberRrucker we need to figure out about the mini panels first.  because if so, any of those `make-md-column()` mixin calls won't work here either. Probably should run this by Mike.

amberRrucker:
@WidgetsBurritos I'm not sure on the mini panel stuff. I just kept the container with the padding and background, because the color should be standard across all quotes and it was already on all of them so I didn't want there to be more things to change in the CMS.

styleguide/derek/_partials/solutions/testimonials.ejs
```diff
@@ -1,15 +1,16 @@
 <!-- Testimonials -->
-<div class="bg-blue-gray standard-padding">
-  <div class="container">
+<div class="quote">
+  <div class="quote-container">
```
WidgetsBurritos: To keep consistent with changes @writes made on other patterns, can use `quote-wrapper` instead of `quote-container`?  Container has a very strict meaning in bootstrap, and while this technically applies in this case, we wan't to avoid any ambiguity.  I realize there's a `quote-wrapper` below, so that would probably need to be named something else.  Perhaps `quote-textWrapper` or something like that.

amberRrucker:
@WidgetsBurritos I've updated this to address your feedback.

amberRrucker:
@runrog Do you mind taking a look at this one as well

WidgetsBurritos:
the only other thing I might add to this is:

Have a look here: 
https://github.com/rackerlabs/zoolander/blob/master/styleguide/derek/solutions/solutions.ejs#L34

See how the other patterns have split up the layouts and the partials?  
For example:
https://github.com/rackerlabs/zoolander/blob/master/styleguide/derek/layouts/business-resources.ejs

It may not be bad idea to do that here too, just for consistency. That technically would help address some of the concerns surrounding the container in the first place. It's not really changing anything other than how it's organized in zoolander, but provides extra clarity as to what should go where in the CMS side

amberRrucker:
Hey @WidgetsBurritos , I addressed the feedback around the partials. Hoping I did what you were asking.

styleguide/derek/_partials/solutions/quote.ejs
```diff
@@ -1,22 +1,12 @@
...
-          <!--Option for logo-->
-          <div class="row red semi-bold half-padding-top">
-            <div class="col-md-2 col-xs-2 center-block">
-              <img src="https://752f77aa107738c25d93-f083e9a6295a3f0714fa019ffdca65c3.ssl.cf1.rackcdn.com/logos/underarmor.png" alt="">
-            </div>
-          </div>
+          <%- partial('../../_partials/cards/quote.ejs', {
+            textWrapperClass: 'quote-split-textWrapper',
+            logoClass: 'quote-split-logo'
+          }) %>
```
WidgetsBurritos: So this is the right idea, but not exactly the organization I had in mind. This code should live in `/styleguide/derek/layouts/quote.ejs` (as it defines layout around the quote), and then the actual quote pattern should live in this page.  

Quotes aren't cards so it doesn't necessarily make sense to move that in there. Make sense?

amberRrucker: @WidgetsBurritos so there are two different quote "patterns" the quote and the split quote. They have two different files. I may just not be super duper clear. We can chat about it in wheaties.

PR #371, opened 2017-02-22
rsweb 8518 adjust hr

amberRrucker:
RSWEB-8518
Adjusting horizontal rule to reduce margins.
Add red option for horizontal rule
Add red border option for background
Add documentation or horizontal rule and red border background 

Viewable here:
http://localhost:9000/derek/assets/swatches

### DIFF
--- a/styleguide/_themes/global/scss/modifiers.scss
+++ b/styleguide/_themes/global/scss/modifiers.scss
@@ -76,6 +76,18 @@ img { max-width: 100%; }
   margin: 0 auto;
 }
 
+// Horizontal rules
+hr {
+  border: 0;
+  border-top: 1px solid $gray-lighter;
+  margin-bottom: $line-height-computed / 4;
+  margin-top:    $line-height-computed / 4;
+}
+
+.hr-red {
+  border-top: 1px solid $rackspace-red;
+}
+
 // Text Alignments
 .center,
 .centered {
@@ -94,6 +106,8 @@ img { max-width: 100%; }
   text-transform: $uppercase;
 }
 
+
+//Backgrounds
 .bg-steel-blue {
   background-color: $blue-steel;
 }
@@ -125,6 +139,11 @@ img { max-width: 100%; }
 .bg-crumpled-paper {
   background-color: $gray-light;
   background-image: url('https://752f77aa107738c25d93-f083e9a6295a3f0714fa019ffdca65c3.ssl.cf1.rackcdn.com/elements/BGcrumpled-paper.jpg');
+}
+
+.bg-red-border {
+  border-bottom: 1px solid $rackspace-red;
+  border-top: 1px solid $rackspace-red;
 }
 
 // Text Colors
--- a/styleguide/_themes/global/scss/scaffolding.scss
+++ b/styleguide/_themes/global/scss/scaffolding.scss
@@ -108,16 +108,6 @@ img {
 }
 
 
-// Horizontal rules
-
-hr {
-  margin-top:    $line-height-computed;
-  margin-bottom: $line-height-computed;
-  border: 0;
-  border-top: 1px solid $hr-border;
-}
-
-
 // Only display content to screen readers
 //
 // See: http://a11yproject.com/posts/how-to-hide-content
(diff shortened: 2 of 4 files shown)


styleguide/_themes/global/scss/modifiers.scss
```diff
@@ -76,6 +76,18 @@ img { max-width: 100%; }
...
 }
 
+// Horizontal rules
+hr {
+  border: 0;
+  border-top: 1px solid $gray-lighter;
+  margin-bottom: $line-height-computed / 2;
+  margin-top:    $line-height-computed / 2;
+
+  &.red {
```
WidgetsBurritos: Since this is new, do we currently have any `<hr class="red">` in use anywhere?   Because it seems like we could easily just have a `hr-red` class that could live separately?

styleguide/_themes/global/scss/modifiers.scss
```diff
@@ -76,6 +76,18 @@ img { max-width: 100%; }
   margin: 0 auto;
 }
 
+// Horizontal rules
+hr {
```
WidgetsBurritos: Looking through the database, we have a crap ton of `<hr>` tags, some with classes, some without.  I think we really should just create a new class for this stuff and then force some of these changes to happen on the admin side, because the implications of just styling all `<hr>` tags, seems like we're going to be perpetuating the same issues we currently have.

amberRrucker: @WidgetsBurritos The hr I'm adding has a class. The only edit there was to the existing hr was to reduce the margin around it, which we decided needed to happen to all the hrs. Since the margin is straight crazy. Would you suggest instead of adding an hr class to add like div class="hr-red" or something? I'm a little wary to start calling things a div with a class of an hr. Since it gets a little complex when we're trying to actually put them on the page.

ejchiefs: I would prefer we use hr tags for this but follow maintainable coding standards in our class naming like @WidgetsBurritos describes.

amberRrucker: I've got it as a classed as hr-red and it remains and hr. 
To further address the feedback I think I would rather treat the disease and not the symptom and get a list of the unclassed hrs and either see why they are there or put them into their own maintainable classes if we think its a good idea. 
I know its a little more work than not adjusting the crazy margins on the hr in general but I think it's ultimately the best option.
@ejchiefs @WidgetsBurritos 

Up to the group.

amberRrucker:
@WidgetsBurritos I've rebased against your branch. JIRA isnt working for me this morning so I can't move the ticket but its done.

styleguide/_themes/global/scss/modifiers.scss
```diff
@@ -76,6 +76,18 @@ img { max-width: 100%; }
...
+  border: 0;
+  border-top: 1px solid $gray-lighter;
+  margin-bottom: $line-height-computed / 4;
+  margin-top:    $line-height-computed / 4;
```
WidgetsBurritos: With the scaffolding changes, can't this now be removed?

styleguide/_themes/global/scss/modifiers.scss
```diff
@@ -76,6 +76,18 @@ img { max-width: 100%; }
...
+
+  &.hr-red {
+    border-top: 1px solid $rackspace-red;
+  }
```
WidgetsBurritos: This should be moved to be a top-level selector.

styleguide/_themes/global/scss/modifiers.scss
amberRrucker: Nope, because then there is nothing defined at all for margin, which means its up to the browser, which isnt always consistent. I would rather set a baseline.

WidgetsBurritos: Then we need to define a class for that and go back to the strategy of people updating the CMS. We're trying to get away from styling tags, not continuing doing it.

amberRrucker: I still don't think that's a good idea. I'd like to maybe have a larger group conversation about it? Can we schedule a wheaties conversation around it?

PR #377, opened 2017-02-23
RSWEB-8507: fix for subnav mobile being cut off on left side

writes:
Fixes the issue when viewing in the mobile viewport where the CTA in the subnav is off the screen on the left side.

http://localhost:9000/derek/incubation/office365

### DIFF
--- a/styleguide/_themes/derek/scss/components/subnav.scss
+++ b/styleguide/_themes/derek/scss/components/subnav.scss
@@ -155,6 +155,7 @@
   }
 
   .subnav-noHover {
+    margin: 5px auto;
     padding-left: 8px;
 
     &:hover {


styleguide/_themes/derek/scss/components/subnav.scss
```diff
@@ -136,7 +136,7 @@
     border-bottom: 0;
     border-left: 2px solid transparent;
     height: auto;
-    margin-top: 0;
+    margin: 0;
```
WidgetsBurritos: I don't think this solves the problem. The issue itself is not with the `.subnav-item` class, but the `.subnav-noHover` class. If you change the following in `.subnav-noHover`, I think it's much cleaner.
```
    margin: 5px 50px 5px 0;
```
to:
```
    margin: 5px auto;
```

writes:
Thank you @WidgetsBurritos for the feedback! I have updated the PR with the suggested changes.

PR #380, opened 2017-02-28
RSWEB-8560: add imac banner image

brantdarilek:


WidgetsBurritos:
That background isn't going to work well with stories that have logos:

e.g. https://www.rackspace.com/customer-stories/feeding-america

brantdarilek:
[run ci]

PR #383, opened 2017-03-02
rsweb-8654-learning-bug

amberRrucker:
RSWEB-8654
Adding back in the background to the learning button to avoid having a green background thats occurring on the /cloud page.


writes:
👍

PR #388, opened 2017-03-07
RSWEB-8591: Add portfolio pattern

WidgetsBurritos:
This PR adds the portfolio pattern into zoolander.

A few notes about this:

1. Mike asked me to not include the icon on the portfolio tabs, so it's just header/text/key offers now.
2. Since this is presently the only acceptable use of vertical tabs in the system, I replaced all existing vertical tab content with portfolio pattern content, instead of adding a new section for product portfolio. I think we covered it last time, but if I need to update any of the usage documentation around that, please let me know. 
3. Chatting with @runrog today, we opted to use gray backgrounds on all of the key offers here. The main issue with this is, without it the content would look oddly indented due to the padding on the key offer pattern. We could theoretically offset this padding with a margin, but then the hover state rests right against the tab line on the left. If CROUX decides to do something else here, that is fine, but thought I'd explain the rational behind the gray background, and also that was after chatting with Roger about it. 
4. While I don't necessarily like it, I did opt to nest `.keyoffersWidget` inside of `.portfolio-wrapper`. The reason for this, is because has a hardcoded background, and it needs to be overridden. This PR didn't seem like the place to try to globally make changes to the key offers pattern, so I opted to limit the impact here by only changing this one context. I'd recommend in the future trying to remove background colors from patterns, and allowing for style modes in the CMS handle that, but that is a task for another day. 
5. While it's never used here in zoolander, I created `.rsTabsHorizontal-contentHeader` to replace the `pattern-header center` class on the horizontal tabs. This can easily be converted and www, and allows us more easily make the shift to maintainable classes here.

### DIFF
--- a/styleguide/_themes/derek/scss/_main.scss
+++ b/styleguide/_themes/derek/scss/_main.scss
@@ -18,6 +18,7 @@
 @import 'patterns/quote';
 @import 'patterns/videoEmbed';
 @import 'patterns/panels';
+@import 'patterns/portfolio';
 @import 'patterns/cards/modules';
 @import 'patterns/cards/keyOffers';
 @import 'patterns/cards/businessResource';
--- a/styleguide/_themes/derek/scss/components/tabs.scss
+++ b/styleguide/_themes/derek/scss/components/tabs.scss
@@ -107,6 +107,14 @@
   }
 }
 
+.rsTabs-contentBody {
+  margin-bottom: 25px;
+}
+
+.rsTabsHorizontal-contentHeader {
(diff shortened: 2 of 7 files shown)


brantdarilek:
LGTM 👍

WidgetsBurritos:
@runrog,

I've updated the spacing between rows. Let me know if that's good for you?

runrog:
+1 As per discussion in CROUX, more space below row on cards. Looks good
screen shot 2017-03-07 at 2 05 05 pm

PR #389, opened 2017-03-07
rsweb-8523-new-corner-pocket

amberRrucker:
RSWEB-8523
Addition of new corner pocket card. Included on solutions pages with documentation and added in the partial of leadSplit to account for the left aligned lead that we use in splits in www.

### DIFF
--- a/styleguide/derek/_partials/cards/corner-pocket.ejs
+++ b/styleguide/derek/_partials/cards/corner-pocket.ejs
@@ -1,0 +1,7 @@
+<div class="cornerPocket-wrapper">
+  <div class="cornerPocket-borderWrapper">
+  	<p class="cornerPocket-heading">Starting at</h3>
+  	<p class="cornerPocket-callout">$0.015/hr</h2>
+  	<p class="cornerPocket-text">High availability without a long-term contract</p>
+	</div>
+</div>
--- a/styleguide/derek/layouts/leadCornerPocket.ejs
+++ b/styleguide/derek/layouts/leadCornerPocket.ejs
@@ -1,0 +1,18 @@
+<div class="container half-padding-full">
+  <div class="row">
+    <div class="col-sm-9">
+      <div class="subpanel">
+        <div class="subpanel-inner">
+          <%- partial('../_partials/solutions/lead.ejs') %>
+        </div>
+      </div>
+    </div>
+    <div class="col-sm-3">
+      <div class="subpanel">
+        <div class="subpanel-inner">
+          <%- partial('../_partials/cards/corner-pocket.ejs') %>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
--- a/styleguide/_themes/derek/scss/_main.scss
+++ b/styleguide/_themes/derek/scss/_main.scss
@@ -19,6 +19,7 @@
 @import 'patterns/videoEmbed';
 @import 'patterns/panels';
 @import 'patterns/cards/modules';
+@import 'patterns/cards/cornerPocket';
 @import 'patterns/cards/keyOffers';
 @import 'patterns/cards/businessResource';
 @import 'patterns/cards/productFeatures';
--- a/styleguide/_themes/derek/scss/patterns/cards/cornerPocket.scss
+++ b/styleguide/_themes/derek/scss/patterns/cards/cornerPocket.scss
@@ -1,0 +1,36 @@
+.cornerPocket-wrapper {
+  background-color: $gray-darkest;
+  color: $white;
+  padding: 15px;
+}
+
+.cornerPocket-borderWrapper {
+  border: 4px solid $white;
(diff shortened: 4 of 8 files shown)


WidgetsBurritos:
Should this text be set to wrap?  It looks weird in some screen sizes. 
image

styleguide/derek/_partials/cards/corner-pocket.ejs
```diff
@@ -0,0 +1,7 @@
+<div class="cornerPocket-wrapper">
+  <div class="cornerPocket-borderWrapper">
+  	<h3 class="cornerPocket-heading">Starting at</h3>
+  	<h2 class="cornerPocket-callout">$0.015/hr</h2>
```
WidgetsBurritos: Do these need to be `<h2>`/`<h3>` tags? I'm not sure this makes semantic sense, as these aren't really headings.

styleguide/derek/_partials/solutions/leadSplit.ejs
```diff
@@ -0,0 +1,13 @@
+<!-- Lead pattern -->
+<!--
+.lead class naming not used in alignment with typical pattern structure
+instead here .lead-wrapper is used to avoid having to override bootstrap base class
+ -->
+<div class="lead-wrapper">
+  <div class="lead-headerWrapper">
+    <h1 class="lead-header left">Title goes here</h1>
```
WidgetsBurritos: Can we have a `lead-headerLeft` instead here? Additionally you may want to confirm this with Mike, because he mentioned something earlier about only ever using leads by themselves at the top of the page, and not inside mini panels like this. He may be okay with it, but I'd just recommend asking him about that.

amberRrucker: We use this with callouts, videos and pdf downloads. Confirming with Mike as we speak. I'd rather not create an all new class since lead already does this in www right?

WidgetsBurritos: Actually, the issue is that it needs to be inside a subpanel well.  This is straight from production: 

image

amberRrucker: Well thats the only way we would use it if its a lead left and something else right would be going into a subpanel well. I dont want to create a class that would only be used for zoolander documentation when we can just drop in the left class that does the same job. @WidgetsBurritos

WidgetsBurritos: You shouldn't need to create the class after all. You can fix this by adjusting your layout to use the appropriate `subpanel` and `subpanel-inner` classes.

styleguide/derek/layouts/leadCornerPocket.ejs
```diff
@@ -0,0 +1,12 @@
+<div class="container half-padding-full">
+  <div class="row">
+    <div class="col-sm-9">
+      <%- partial('../_partials/solutions/leadSplit.ejs') %>
```
WidgetsBurritos: Surround this with:
```
<div class="subpanel">
  <div class="subpanel-inner">
    ...
  </div>
</div>
```

styleguide/derek/layouts/leadCornerPocket.ejs
```diff
@@ -0,0 +1,12 @@
...
+    </div>
+    <div class="col-sm-3">
+      <%- partial('../_partials/cards/corner-pocket.ejs', {
+        type: 'video'
+      }) %>
```
WidgetsBurritos: Same here

amberRrucker:
@WidgetsBurritos Fixed It!

styleguide/derek/layouts/leadCornerPocket.ejs
```diff
@@ -0,0 +1,16 @@
...
+        </div>
+        <div class="col-sm-3">
+          <%- partial('../_partials/cards/corner-pocket.ejs', {
+            type: 'video'
+          }) %>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
```
WidgetsBurritos: Close, but the subpanels need to be inside the individual columns, not outside of it. 

```
<div class="container half-padding-full">
  <div class="row">
    <div class="col-sm-9">
      <div class="subpanel">
        <div class="subpanel-inner">
          <%- partial('../_partials/solutions/lead.ejs') %>
        </div>
      </div>
    </div>
    <div class="col-sm-3">
      <div class="subpanel">
        <div class="subpanel-inner">
          <%- partial('../_partials/cards/corner-pocket.ejs', {
            type: 'video'
          }) %>
        </div>
      </div>
    </div>
  </div>
</div>
```

amberRrucker: Ok well then I lied but its fixed now @WidgetsBurritos 
Thank you :/

PR #393, opened 2017-03-15
rsweb-8689-app-style-icons

amberRrucker:
RSWEB-8689

Reproducing the "app style" icons on mobile in the ceiling.
Removing the text, increasing the size of the icons.

### DIFF
--- a/styleguide/_themes/derek/scss/globalElements/mainnav.scss
+++ b/styleguide/_themes/derek/scss/globalElements/mainnav.scss
@@ -284,6 +284,19 @@
     width: 95px;
   }
 
+  .navbar-phoneicon,
+  .navbar-chaticon,
+  .navbar-emailicon,
+  .navbar-searchicon {
+
+    &::before {
+      font-size: 1.85rem;
+      line-height: 2rem;
+      margin-right: 0;
+      padding: 0 23px;
+    }
+  }
+
   .main-nav {
 
     .navbar-hamburger {
@@ -292,16 +305,6 @@
         content: '\f0c9';
         display: inline-block;
         margin: 17px 10px;
-      }
-    }
-
-    .navbar-searchicon {
-      float: left;
-
-      &::after {
-        font-size: 5vw;
-        line-height: 1vw;
-        margin: 0;
       }
     }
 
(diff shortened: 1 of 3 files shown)


styleguide/_themes/derek/scss/globalElements/mainnav.scss
```diff
@@ -344,20 +344,19 @@
...
 
-  .main-nav {
-
-    .navbar-phoneicon,
-    .navbar-chaticon,
-    .navbar-emailicon,
-    .navbar-searchicon {
-      @include make-xs-column(2);
-      border-left: solid 1px $gray-darker;
-      height: 63px;
```
amberRrucker: this was unnecessary code that I didn't clean out when sexy nav was being developed.

styleguide/derek/_partials/_ceiling.ejs
```diff
@@ -5,13 +5,13 @@
           <nav class="ceiling-menuNav">
               <ul class="ceiling-menuInline" role="menu">
                 <li class="ceiling-menuInline-item">
-                  <a class="navbar-phoneicon" title="Sales/Support Call"  data-toggle="modal" data-target="#myModal">Call Us</a>
+                  <a class="navbar-phoneicon" title="Sales/Support Call"  data-toggle="modal" data-target="#myModal"><span class="hidden-xs">Call Us</span></a>
```
amberRrucker: Thinking @WidgetsBurritos is probably going to want this with a more maintainable class name huh?

WidgetsBurritos: We're already using `hidden-xs` further down the file, so I'm not too worried about it here. If we ever refactor this in the future, then maybe we can deal with that then for now, I'm okay with this.

runrog:
@amberRrucker The only weird thing here is the email envelope. It's technically centered in it's div but that extra space is making it look off.. I added a border on the right? maybe bring in some of that right padding? 
screen shot 2017-03-16 at 10 12 29 am

amberRrucker:
@runrog good point. I'll do that real quick like.

amberRrucker:
@runrog fixed the alignment issue.

PR #394, opened 2017-03-15
rsweb-8714-reduce-nav-font-size

amberRrucker:
RSWEB-8714

Decreasing font size and adding partners into the navigation
Can be viewed on http://localhost:9000/derek/incubation/homepage
Please open search to test breakpoints fully

### DIFF
--- a/styleguide/_themes/derek/scss/globalElements/ceiling.scss
+++ b/styleguide/_themes/derek/scss/globalElements/ceiling.scss
@@ -1,8 +1,8 @@
 // see styleguide/_themes/global/scss/_bootstrap_overrides.scss for sass lint overrides.
 
 $ceiling-margin: floor($navbar-margin-bottom / 4);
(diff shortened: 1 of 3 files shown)


runrog:
found 1 thing: 
The top right nav buttons fonts are also smaller and the ones without icons look misaligned, may be that you need to also shrink the icon font size as well
screen shot 2017-03-16 at 10 30 56 am

runrog:
my bad they all have icons* They do look misaligned though. Other than that, the rest looks good

amberRrucker:
@runrog good catch. Adjusted that.

amberRrucker:
Hey @WidgetsBurritos in order to solve for screen sizes smaller than 380px i had to add a media query for super small. This will actually end up being solved when we merge in the "app style" icons for mobile so we can remove this on that rebase.

PR #397, opened 2017-03-17
RSWEB-8753: added new filter side bar, removed teal from events

runrog:
Note: If styling is merged, it will break the current events page styling. This will need to wait until www work is done on the filtering. We'll also need a www PR for the js :/

Zoolander Link:
http://localhost:9000/derek/incubation/events


amberRrucker:
screen shot 2017-03-17 at 2 57 22 pm
screen shot 2017-03-17 at 2 57 42 pm

two things:
can we add some padding to the right side of the location
can we adjust the alignment so that the date and title align on the left

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,127 @@ jQuery(document).ready(function($) {
...
+    });
+
+    // Set a slight delay to the resize functionality so that it doesn't
+    // run on every pixel of a browser resize
+    var resizeTimer;
+    $rsWindow.resize(function(){
+      clearTimeout(resizeTimer);
+      resizeTimer = setTimeout(function() {
+        // If not mobile
+        if ( $rsWindow.outerWidth() > 768 ) {
```
WidgetsBurritos: check this number

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,127 @@ jQuery(document).ready(function($) {
...
+    function checkInBounds(wt) {
+      if ($stickyItemTop - $mainNavHeight < wt &&
+        $filterBar.outerHeight() + $filterBar.offset().top +
+        $mainNavHeight - ($filterForm.outerHeight() * 3) >
+        (wt + $filterForm.outerHeight())) {
+          inBounds = true;
+      } else {
+        inBounds = false;
+      }
+      return inBounds;
```
WidgetsBurritos: If this condition is just setting a boolean, you can just do: 
```
return ($stickyItemTop - $mainNavHeight < wt &&
         $filterBar.outerHeight() + $filterBar.offset().top +
         $mainNavHeight - ($filterForm.outerHeight() * 3) >
         (wt + $filterForm.outerHeight()));

File: styleguide/js/zoolander.js [hunk 3]
```diff
@@ -126,6 +128,127 @@ jQuery(document).ready(function($) {
...
+    }
+
+    $filterHamburger.click(function(){
```
WidgetsBurritos: can you add a space after ()

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,127 @@ jQuery(document).ready(function($) {
...
+          });
+          isFixed = false;
+        }
+      }
+    });
+
+    // Set a slight delay to the resize functionality so that it doesn't
+    // run on every pixel of a browser resize
+    var resizeTimer;
+    $rsWindow.resize(function(){
```
WidgetsBurritos: space after ()

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,127 @@ jQuery(document).ready(function($) {
...
+              'width' : ''
+            });
+            isFixed = false;
+          }
+        }
+        // Make sure form is always shown in appropriate viewport
+        // given the users interactions with browser resize.
+        if (!$filterForm.is(':visible') && $rsWindow.outerWidth() > 767) {
+          $filterForm.show();
+        } else if ($filterForm.is(':visible') && $rsWindow.outerWidth() < 768) {
```
WidgetsBurritos: check these breakpoints too.

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,127 @@ jQuery(document).ready(function($) {
...
+      }, 250);
+    });
+
+    // for zoolander [since side menu auto closes,
+    // it gets wrong initial width]. Will not be in prod
+    setTimeout(function(){
+      $filterForm.css({
+        'width' : getWidth($filterBar, $barPadding) + 'px'
+      });
+    },1000)
```
WidgetsBurritos: you should be fine to scrap this

styleguide/_themes/derek/scss/snowflakes/filtering.scss
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+.rsFilter-title-mobile {
+  @include subhead($gray-darker);
+  display: none;
+  text-align: center;
+}
+
+.rsFilter-form-title {
```
WidgetsBurritos: can we switch to camel case after the first hyphen?

styleguide/_themes/derek/scss/snowflakes/filtering.scss
```diff
@@ -0,0 +1,94 @@
...
+  &:hover {
+    background-color: $gray-light;
+  }
+
+  &:focus {
+    outline: 0;
+  }
+}
+
+.rsFilter-title-mobile {
```
WidgetsBurritos: to match below, can we swap this to `mobileTitle` instead?

styleguide/_themes/derek/scss/snowflakes/filtering.scss
```diff
@@ -0,0 +1,94 @@
...
+  display: none;
+  text-align: center;
+}
+
+.rsFilter-form-title {
+  @include subhead($gray-darker);
+  text-align: center;
+}
+
+.rsFilter-form-select {
```
WidgetsBurritos: camelCase here too?

styleguide/_themes/derek/scss/snowflakes/filtering.scss
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+.rsFilter-content {
+  @include make-sm-column(10);
+  padding: 25px;
+}
+
+// Responsive behavior
+@media (max-width: $screen-sm-min) {
```
WidgetsBurritos: Check this breakpoint.  max-widths should technically match `*-max` variables (although this isn't always the case), but general rule of thumb here.  Could be related to the break point issues.

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,127 @@ jQuery(document).ready(function($) {
...
+      return inBounds;
+    }
+
+    $filterHamburger.click(function(){
+      $filterForm.slideToggle(150);
+    });
+
+    $rsWindow.scroll(function() {
+      // If not in mobile
+      if ($rsWindow.outerWidth() > 767) {
```
WidgetsBurritos: here too

styleguide/derek/incubation/events.ejs
```diff
@@ -19,27 +19,29 @@
...
-                <option value="5">Tradeshow/Conference</option>
-                <option value="6">Training</option>
-                <option value="7">Webinar</option>
+  <div class="rsFilter-section">
+    <div class="rsFilter-row row-eq-height">
+      <div class="rsFilter-bar">
+        <button class="rsFilter-hamburger"></button>
+        <div class="rsFilter-title-mobile">Find a Story</div>
+        <form class="rsFilter-form">
+          <div class="rsFilter-form-title">Find a Story</div>
```
WidgetsBurritos: This should say "Event"

styleguide/derek/incubation/events.ejs
```diff
@@ -19,27 +19,29 @@
...
-                <option value="3">Field Events</option>
-                <option value="4">Forum</option>
-                <option value="5">Tradeshow/Conference</option>
-                <option value="6">Training</option>
-                <option value="7">Webinar</option>
+  <div class="rsFilter-section">
+    <div class="rsFilter-row row-eq-height">
+      <div class="rsFilter-bar">
+        <button class="rsFilter-hamburger"></button>
+        <div class="rsFilter-title-mobile">Find a Story</div>
```
WidgetsBurritos: This should say "Event"

runrog:
@WidgetsBurritos finished tweaks, including mobile scss first aproach. I adjusted one media query check in the js. Not seeing the hamburger overlap any more.

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,114 @@ jQuery(document).ready(function($) {
...
+            isFixed = false;
+          }
+        }
+        // Make sure form is always shown in appropriate viewport
+        // given the users interactions with browser resize.
+        if (!$filterForm.is(':visible') && $rsWindow.outerWidth() > 768) {
+          $filterForm.show();
+        } else if ($filterForm.is(':visible') && $rsWindow.outerWidth() < 768) {
+          $filterForm.hide();
+        }
```
WidgetsBurritos: Is this hide/show functionality here even necessary? Seems like this CSS controls this behavior already: 

```
.rsFilter-form {
  display: none;
}

  .rsFilter-form {
    display: block; 
  }
}
```

Additionally by letting css handle it, instead of inside this 250ms delayed JS, we no longer have that random glitch where the sidebar just randomly disappears.

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,114 @@ jQuery(document).ready(function($) {
...
+      }
+    });
+
+    // Set a slight delay to the resize functionality so that it doesn't
+    // run on every pixel of a browser resize
+    var resizeTimer;
+    $rsWindow.resize(function() {
+      clearTimeout(resizeTimer);
+      resizeTimer = setTimeout(function() {
+      var $rsWindowTop = $rsWindow.scrollTop();
```
WidgetsBurritos: Can you fix this indention here?

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,114 @@ jQuery(document).ready(function($) {
...
+            $filterForm.css({
+              'position' : 'fixed',
+              'width' : getWidth($filterBar, $barPadding) + 'px'
+            });
+            isFixed = true;
+          } else if (isFixed && inBounds($rsWindowTop)) {
+            $filterForm.css({
+              'width' : getWidth($filterBar, $barPadding) + 'px'
+            });
+          } else if (isFixed && !inBounds($rsWindowTop)) {
```
WidgetsBurritos: `inBounds()` isn't exactly the lightest of functions. It can potentially run three times here, when it really only needs to be ran once per function call. Can we store this value into a variable prior to these conditions to save the unnecessary loads? 

```
var isInBounds = inBounds($rsWindowTop);
if (!isFixed && isInBounds) {
  ...
} else if (isFixed && isInBounds) {
  ...
} else if (isFixed && isInBounds) {
  ...
}
```

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,114 @@ jQuery(document).ready(function($) {
...
+      var width = parseInt(bar.outerWidth(true), 10) - padding;
+      return width;
+    }
+
+    // Callback to check if in bounds of side menu
+    function inBounds(wt) {
+      return $stickyItemTop - $mainNavHeight < wt &&
+        $filterBar.outerHeight() + $filterBar.offset().top +
+        $mainNavHeight - ($filterForm.outerHeight() * 3) >
+        (wt + $filterForm.outerHeight())
```
WidgetsBurritos: So I know I previously recommended just flat out returning this, but now that I see it, I might recommend tweaking it a little bit more, as it's a bit confusing as to what's going on here. This is a rather long condition and doesn't really tell us a whole lot of what is actually going on.  Perhaps something like this would be better?
```
    function inBounds(wt) {
      var filterFormHeight = $filterForm.outerHeight();
      var windowTopInBounds = $stickyItemTop - $mainNavHeight < wt;
      var filterBottom = $filterBar.outerHeight() + $filterBar.offset().top;
      var formBottom = wt + filterFormHeight;
      var filterInBounds = filterBottom + $mainNavHeight - filterFormHeight * 3 > formBottom;
      return windowTopInBounds && filterInBounds;
    }
```
Something like this feels a little bit more expressive in my opinion, and won't have us scratching our heads 6 months down the road trying to figure out what was going on here. 

Additionally, this is a prime example of why it would be good to have JS unit testing setup in zoolander.  Whenever testing does get added, I'd recommend this function be one of the first things to get tested.

styleguide/js/zoolander.js
runrog: this is why I had the way I did before. one function, and its stored a variable

runrog: also it can't just be stored once, you have to keep checking if you are in bounds of the side bar when you're scrolling up and down the page. It's not always going to be the same value.

WidgetsBurritos: I know.  My proposed change above, only affects it inside this function. So yes it's gonna rerun, but it's only gonna run once per function call, not potentially 3 times per function call.

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,118 @@ jQuery(document).ready(function($) {
...
+      var filterInBounds = filterBottom + $mainNavHeight - filterFormHeight * 3 > formBottom;
+      inBounds = windowTopInBounds && filterInBounds;
+      return inBounds;
+    }
+
+    $filterHamburger.click(function() {
+      $filterForm.toggleClass('rsFilter-mobileForm');
+      setTimeout(function() {
+        $filterForm.toggleClass('rsFilter-mobileForm-effect');
+      }, 100)
```
WidgetsBurritos: semi colon

styleguide/js/zoolander.js
```diff
@@ -126,6 +128,118 @@ jQuery(document).ready(function($) {
...
+            isFixed = true;
+          } else if (isFixed && inBounds) {
+            $filterForm.css({
+              'width' : getWidth($filterBar, $barPadding) + 'px'
+            });
+          } else if (isFixed && !inBounds) {
+            $filterForm.css({
+              'position' : ''
+            });
+            isFixed = false
```
WidgetsBurritos: semi colon

WidgetsBurritos:

PR #398, opened 2017-03-22
RSWEB-8705: consistent gutters

amberRrucker:
RSWEB-8705 
Getting the gutters into some semblance of normal.

You may notice some additional padding on the rug, footer, basement and imac banner containers. In order to have those elements align with the beginning of the actual word of the links in the main navigation and subnav instead of the beginning of the padded link container (you can see this on those links hover states) those containers got some extra padding. The actual first word of those links also now match up with the left side of the page elements, which is why I didnt instead remove padding from the main nav and subnav. 

Should only need to remove the container declaration from the ceiling in www

This can best be seen on this page:
http://localhost:9000/derek/templates/section-overview

### DIFF
--- a/styleguide/_themes/global/scss/banners.scss
+++ b/styleguide/_themes/global/scss/banners.scss
@@ -22,10 +22,7 @@
 .imacBanner-container {
   @include make-container;
   @include fadein(1s);
-}
-
-.imacBanner-textRow {
-  @include make-row;
+  padding: 0 45px;
 }
 
 .imacBanner-headlineLarge {
@@ -54,6 +51,11 @@
   margin: 0 auto;
   padding: 15px;
   position: relative;
+}
+
+.imacBanner-textRow {
+  @include make-row;
+  padding: 0 25px;
 }
 
 .imacBanner-buttonRow {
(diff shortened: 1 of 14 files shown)


styleguide/_themes/derek/scss/globalElements/basement.scss
```diff
@@ -21,7 +21,7 @@
 
 .basement-container {
   @include make-container;
-  padding: 10px 0;
+  padding: 10px 25px;
```
amberRrucker: putting back to standard container padding but still has to 10px top and bottom

WidgetsBurritos:
So we didn't catch it before, but since this is about gutters apparently [this change happened],
 and broke the gutter related to the subnav button.
 (https://github.com/rackerlabs/zoolander/pull/387/files#diff-09f2c53c9e8ff2d3b25453b36e5581edL111): 
```
.subnav-noHover {
  margin: 5px $grid-gutter-width 5px 0;
  padding: 8px 0;
}
```
was changed to: 
```
.subnav-noHover {
  float: right;
  margin-top: 1px;
}
```

That right margin was important gutter-wise because that was what prevented this from happening:
image

Given that this task is related to fixing margins, this seems like as good as place as any to fix it.

WidgetsBurritos:
This has side scrolling issues from 1280-1304px. Appears to be related to `half-padding-full` on the outermost container.

image

amberRrucker:
@WidgetsBurritos Ive addressed the feedback you have except for the button margin because it was decided in the croux that we're hoping to get the chat stuff pushed to the bottom and since it really has minimal impact cases then its totally okay.

WidgetsBurritos:

WidgetsBurritos:
[run ci]

PR #381, opened 2017-03-02
Rsweb 8496 backtotop

amberRrucker:
RSWEB-8496
Back to top Improvements:
- Removing back to top from desktop view
- Refactoring back to top pattern
- Making sure back to top pattern is included on zoolander pages
- Switching background color for back to top
- Adjusting padding on back to top pattern

Viewable on all template and incubation pages in mobile.

PR #403, opened 2017-03-28
RSWEB-8876: Cleanup Javascript + Integrate Travis

WidgetsBurritos:
This PR cleans up bad javascript and integrates travis for javascript linting.

### DIFF
--- a/styleguide/global/global.js
+++ b/styleguide/global/global.js
@@ -1,199 +1,172 @@
-jQuery(document).ready(function($) {
-  //main nav functionality
-  var $topLevelDropDown = $('.navbar-topItem > .navbar-topLink-dropDown');
-  var $navBurger = $(".navbar-hamburger");
+jQuery(document).ready(($) => {
+  // main nav functionality
+  const $topLevelDropDown = $('.navbar-topItem > .navbar-topLink-dropDown');
+  const $navBurger = $('.navbar-hamburger');
 
-  $navBurger.click(function(){
-    $navBurger.toggleClass("collapsed");
+  $navBurger.click((e) => {
+    $navBurger.toggleClass('collapsed');
 
-    //reset dopdown carrets if collapsing the menu
-    if($(this).hasClass('collapsed')){
+    // reset dopdown carrets if collapsing the menu
+    if ($(e.currentTarget).hasClass('collapsed')) {
       $topLevelDropDown.removeClass('navbar-dropDown-triggerActive');
     }
   });
 
-  //Toggle the open arrows on, top level, non drop down links
-  $('.navbar-topLink:not(.navbar-topLink.navbar-topLink-dropDown)').click(function(){
-    var $el = $(this).parent();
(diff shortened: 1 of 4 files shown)


styleguide/global/global.js
```diff
@@ -1,199 +1,172 @@
-jQuery(document).ready(function($) {
-  //main nav functionality
-  var $topLevelDropDown = $('.navbar-topItem > .navbar-topLink-dropDown');
-  var $navBurger = $(".navbar-hamburger");
+jQuery(document).ready(($) => {
+  // main nav functionality
+  const $topLevelDropDown = $('.navbar-topItem > .navbar-topLink-dropDown');
+  const $navBurger = $('.navbar-hamburger');
```
amberRrucker: bless you arrow functions. bless you.

styleguide/global/global.js
```diff
@@ -1,199 +1,172 @@
...
-    return false;
-  });
-  jQuery (window).scroll (function () {
-    if (jQuery (this).scrollTop () > offset) {
-      jQuery ('.back-to-top').fadeIn (duration);
-    } else {
-      jQuery ('.back-to-top').fadeOut (duration);
-    }
-  });
-
```
amberRrucker: Hey guys, were still using back to top. Its just limited to mobile now.

styleguide/global/global.js
```diff
@@ -1,199 +1,172 @@
...
-    jQuery ('html, body').animate ({scrollTop: 0}, duration);
-    return false;
-  });
-  jQuery (window).scroll (function () {
-    if (jQuery (this).scrollTop () > offset) {
-      jQuery ('.back-to-top').fadeIn (duration);
-    } else {
-      jQuery ('.back-to-top').fadeOut (duration);
-    }
-  });
```
WidgetsBurritos: stripped all this because it looked like we weren't using it any more.  @amberRrucker is this correct?

amberRrucker: Nosiree. Just let a review comment on it. We're still using it in mobile.

WidgetsBurritos: @amberRrucker 
It's just using standard anchors for that functionality.  You got rid of references to the `back-to-top` class here: 
https://github.com/rackerlabs/zoolander/pull/381/files#diff-eac2bdcca34b7ff9557618cf91617413L9

WidgetsBurritos: Furthermore this javascript doesn't even exist in www

amberRrucker: Yeah so the js has actually been broken on the scrolling effect for awhile and I wasnt sure why but there was a bug in to fix it. I think if we could fix it instead of removing it that would be super cool.

WidgetsBurritos: @amberRrucker I added back the click event back (and changed the trigger class).  However the fading in/out based on vertical position doesn't make sense anymore with the new design so I left that portion off.  Note: this still isn't being pulled into www.  That will need to be handled separately.

PR #405, opened 2017-03-29
RSWEB-8848: Add live reload

xinbin:
https://jira.rax.io/browse/RSWEB-8848

Change summary:
* Add browser sync
* Add script for starting browser sync
* Add browser sync snip for injecting updates

Instructions:
1. npm install
2. npm run babel
3. npm start
4. npm bs-start
5. navigate to http://localhost:3000/derek/templates/landing-page/video-LP
6. go edit styleguide/_themes/derek/scss/patterns/quote.scss , save
7. view your browser again .. should update
8. also go look at your terminal where you're running browsersync, you should see:
1__npm_run_bs-start__node_

- - - - - - - 
To be discussed:
do we want to implement it this way with the snip added?
what files do we want to watch?
do we want browser sync to start when you start harp, or keep it a separate command?

### DIFF
--- a/bs-config.js
+++ b/bs-config.js
@@ -1,0 +1,17 @@
+/*
+ |--------------------------------------------------------------------------
+ | Browser-sync config file
+ |--------------------------------------------------------------------------
+ |
+ | For up-to-date information about the options:
+ |   http://www.browsersync.io/docs/options/
+ |
+ | There are more options than you see here, these are just the ones that are
+ | set internally. See the website for more info.
+ |
+ |
+ */
+module.exports = {
+  files: ['styleguide/js/**/**/*.js', 'styleguide/_themes/**/**/*.scss'],
+  proxy: 'localhost:9000',
+};
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "scripts": {
     "harp": "harp ",
     "harp:compile": "npm run harp compile styleguide dist",
-    "start": "harp server styleguide",
+    "start": "concurrently 'harp server styleguide' 'browser-sync start --config bs-config.js'",
     "gh-pages": "./scripts/build-pages.sh",
     "preversion": "npm run babel && npm run harp compile styleguide dist && git add -f dist",
     "postversion": "git push && git push --tags",
@@ -24,11 +24,13 @@
   "license": "MIT",
   "dependencies": {
     "bootstrap-sass": "^3.3.6",
+    "concurrently": "^3.4.0",
     "harp": "^0.20.1"
   },
   "devDependencies": {
     "babel-cli": "^6.24.0",
     "babel-preset-env": "^1.2.2",
+    "browser-sync": "^2.18.8",
     "eslint": "^3.18.0",
     "eslint-config-airbnb-base": "^11.1.2",
     "eslint-plugin-import": "^2.2.0",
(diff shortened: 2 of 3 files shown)


WidgetsBurritos:

WidgetsBurritos:

WidgetsBurritos:
[run ci]

WidgetsBurritos:
[run ci]

bs-config.js
```diff
@@ -0,0 +1,17 @@
...
+ | For up-to-date information about the options:
+ |   http://www.browsersync.io/docs/options/
+ |
+ | There are more options than you see here, these are just the ones that are
+ | set internally. See the website for more info.
+ |
+ |
+ */
+module.exports = {
+  files: ['styleguide/_themes/derek/scss/**/*.scss', 'styleguide/_themes/**/**/*.scss'],
```
xinbin: Need to flesh out what files / paths to add here.

styleguide/_layouts/default.jade
```diff
@@ -22,64 +22,65 @@ html(lang="en")
     script(src="https://code.jquery.com/jquery-1.12.3.min.js" integrity="sha256-aaODHAgvwQW1bFOGXMeX+pC4PZIPsvn2h1sArYOhgXQ=" crossorigin="anonymous")
     script(src="https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/zoolander/modernizr.min.js")
     script(src="https://577351e10bc4a0198d93-f60a0bb748a3c84145bb10da7563bafb.ssl.cf1.rackcdn.com/zoolander/jquery.fittext.js")
+  
+  body
```
xinbin: What I really added was the body tag.. browser sync needs the body tag

WidgetsBurritos:
@xinbin @bighappyface @runrog 
Would it make sense to install something like [concurrently](https://www.npmjs.com/package/concurrently) here, and then we could just modify the start command to: 
```
"start": "concurrently 'harp server styleguide' 'browser-sync start --config bs-config.js'",
```

That way `npm start` runs both browser-sync and harp simultaneously?

WidgetsBurritos:
Apart from my feedback above, this is looking good. We'll probably need to talk through the particular files we want to watch tomorrow in wheaties. It seems like babel should watch JS, and maybe sass should watch scss files, and then we can just have browser sync watch the processed js/css files, but I haven't fully thought that through.

xinbin:
I think concurrently makes sense also: [why concurrently](https://www.npmjs.com/package/concurrently#why)
down for that

bs-config.js
WidgetsBurritos: Now that we have babel watch setup, can we add these paths? 
`styleguide/js/*.js`
`styleguide/js/*/*/*.js

package.json
```diff
@@ -8,7 +8,8 @@
   "scripts": {
     "harp": "harp ",
     "harp:compile": "npm run harp compile styleguide dist",
-    "start": "harp server styleguide",
+    "start": "concurrently 'harp server styleguide' 'browser-sync start --config bs-config.js'",
+    "bs-start": "browser-sync start --config bs-config.js",
```
WidgetsBurritos: if this is included inside of `start` is this still necessary? I'm okay with keeping it, but if we do decide to keep it, maybe having a `npm harp-start` command makes sense as well, to have the facilities to start one at a time if we need to.  I don't mind either way, I just don't think it makes sense to have a browser-sync only start command, if we don't have one for harp.

xinbin: was only keeping it in case someone found it useful :) ,but yes it is not what I would consider needed. I'll remove it. we can add something like that later if it becomes necessary.

WidgetsBurritos: sounds good

PR #408, opened 2017-03-29
RSWEB-8843: modularized rug as global component

runrog:
You can see basement on the following zoolander page:


brantdarilek:
👍

PR #410, opened 2017-03-30
RSWEB-8843: modularized ceiling as global component

runrog:
Can see homepage ceiling here

Landing Page ceiling here
http://localhost:9000/derek/templates/landing-page/cta-LP


brantdarilek:
👍

PR #411, opened 2017-03-30
RSWEB-8843: modularized subnav as global component

runrog:
Subnav can be seen here
http://localhost:9000/derek/incubation/office365


brantdarilek:
👍

PR #413, opened 2017-03-30
RSWEB-8843: feat(zoolander) modularized nav & regionSelect as global components

runrog:
Can see nav here

Click region selector in basement to see the modal


WidgetsBurritos:
@runrog can you please make the branch name be all lower case and push a separate PR? The PR builder doesn't work with capital letters right now.

PR #424, opened 2017-04-03
RSWEB-6759: add currency selector

writes:
Adds currency selector styles and markup to basement template for consistency across web properties.

http://rsweb-6759-currency-basement-selector.www.dev.website.rackspace.com/openstack/public/service-levels

### DIFF
--- a/styleguide/derek/global-components/basement/_basement.scss
+++ b/styleguide/derek/global-components/basement/_basement.scss
@@ -20,6 +20,31 @@
   text-align: center;
 }
 
+.basement-currency,
+.region-select {
+  @include make-sm-column(6);
+  text-align: center;
+}
+
+.basement-currency {
+  @include make-lg-column-pull(1);
+  padding: 0;
+}
+
+.basement-currency-select {
+  background-color: $white;
+}
+
+.basement-currency-select,
+.basement-currency-text {
+  font-size: 1.2rem;
+}
+
+.basement-currency-text {
+  color: $white;
+  text-transform: uppercase;
+}
+
 .basementLp-copyright {
   color: $gray-light;
   margin-top: 5px;
@@ -121,7 +146,7 @@
 }
 
 .region-select {
-  text-align: center;
+  @include make-md-column(4);
 }
 
 @media screen and (min-width: $screen-sm-min) {
@@ -129,6 +154,7 @@
     text-align: right;
   }
 
+  .basement-currency,
   .region-select {
     text-align: left;
   }
--- a/styleguide/derek/global-components/basement/basement.ejs
+++ b/styleguide/derek/global-components/basement/basement.ejs
@@ -1,15 +1,24 @@
 <!--[basement]-->
 <div class="bg-dark-gray">
   <div class="basement-container">
-    <div class="col-sm-4">
+    <div class="col-sm-9">
       <div class="region-select">
         <span class="uppercase white basement-region-select">Change your region:</span>
         <a class="basement-flag-link" data-toggle="modal" id="region" data-target="#regionSelect">
           <div class="basement-flag-icon basement-flag-uk"></div>
         </a>
       </div>
+      <div class="basement-currency">
+        <span class="currency_text basement-currency-text">Currency:</span>
+        <select class="currency_selector basement-currency-select">
+          <option value="USD">USD</option>
+          <option value="GBP">GBP</option>
+          <option value="AUD">AUD</option>
+          <option value="EUR">EUR</option>
+        </select>
+      </div>
     </div>
-    <div class="col-sm-3 col-sm-push-5">
+    <div class="col-sm-3">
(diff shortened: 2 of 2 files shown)


WidgetsBurritos:

WidgetsBurritos:
ok to test

WidgetsBurritos:
add to whitelist

WidgetsBurritos:
Grant it, the ticket is 9 months old, but the ticket says `The currency selector in the basement is not styled and should sit next to the region selector, not below.`  

I'm not sure if that is still relevant or not, but it's worth noting it still is showing up below, not next to, the region selector. 
image

writes:
Thanks @WidgetsBurritos I totally missed that part.

styleguide/derek/global-components/basement/basement.ejs
```diff
@@ -1,15 +1,24 @@
...
   <div class="basement-container">
-    <div class="col-sm-4">
+    <div class="col-sm-9">
       <div class="region-select">
         <span class="uppercase white basement-region-select">Change your region:</span>
         <a class="basement-flag-link" data-toggle="modal" id="region" data-target="#regionSelect">
           <div class="basement-flag-icon basement-flag-uk"></div>
         </a>
       </div>
+      <div class="basement-currency col-lg-pull-1">
```
WidgetsBurritos: If you're gonna do this, probably could just add: `@include make-lg-column-pull(1);` to the definition of basement-currency.

WidgetsBurritos: From a zoolander perspective, we need to be able to illustrate (and more importantly test) the basement both with and without a pricing selector. 

I'd recommend making this conditional using jade.
Basically the idea would be 
```
<% if(!hide_currency_selector) { var hide_currency_selector = false; } %>
```

And then wrap this code with something like this:
```
<% if (!hide_currency_selector) { %>
  Your code here
<% } %>
```

Then in references to the basement.ejs file, we set `hide_currency_selector` to `true`, except for one particular instance.

styleguide/derek/global-components/basement/basement.ejs
```diff
@@ -1,15 +1,28 @@
...
   <div class="basement-container">
-    <div class="col-sm-4">
+    <div class="col-sm-9">
       <div class="region-select">
         <span class="uppercase white basement-region-select">Change your region:</span>
         <a class="basement-flag-link" data-toggle="modal" id="region" data-target="#regionSelect">
           <div class="basement-flag-icon basement-flag-uk"></div>
         </a>
       </div>
+      <% if (!hide_currency_selector) { %>
```
WidgetsBurritos: This should be:
```
<% if (hide_currency_selector) { %>
```
instead of:
```
<% if (!hide_currency_selector) { %>
```

styleguide/derek/global-components/basement/_basement.scss
```diff
@@ -29,6 +29,34 @@
...
+  text-align: center;
+}
+
+// Override for bootstrap padding
+div {
+  &.basement-currency {
+    @include make-lg-column-pull(1);
+    padding: 0;
+  }
+}
```
WidgetsBurritos: Why is this div needed here? Does this not suffice? 
```
.basement-currency {
  @include make-lg-column-pull(1);
  padding: 0;
}
```

When I test it locally it seems fine to me, unless I'm missing something.

styleguide/derek/global-components/basement/basement.ejs
```diff
@@ -1,15 +1,28 @@
+<% if(!hide_currency_selector) { var hide_currency_selector = true; } %>
```
WidgetsBurritos: We're still missing one part to make this work.  We need one of the files to reference this.  As it's kind of trivial which page has this, I'd recommend just changing the the `/derek/incubation/office365.ejs` file to:

```
<%- partial("../_partials/_basement", {hide_currency_selector: false}) %>
```
That said, if @amberRrucker would prefer us having it on another page instead, we should double check with her on that.

WidgetsBurritos:
[run ci]

styleguide/derek/incubation/office365.ejs
```diff
@@ -2693,4 +2693,4 @@
 
 <%- partial("../_partials/_rug") %>
 <%- partial("../_partials/_footer") %>
-<%- partial("../_partials/_basement") %>
+<%- partial("../_partials/_basement", {show_currency_selector: true}) %>
```
WidgetsBurritos: kill this

styleguide/derek/global-components/basement/basement.ejs
```diff
@@ -1,15 +1,28 @@
+<% if(!show_currency_selector) { var show_currency_selector = false; } %>
```
WidgetsBurritos: kill all ejs references here

PR #426, opened 2017-04-04
RSWEB-8819: [zoolander] Integrate Javascript unit testing

xinbin:
https://jira.rax.io/browse/RSWEB-8819

Change Summary:
* Add in Karma 
* Uses mocha, phantomjs, and firefox

---- -- --- - - - -

1. npm install
2. karma start

You'll see a failed test:
karma_run


WidgetsBurritos:

WidgetsBurritos:
ok to test

WidgetsBurritos:
add to whitelist

package.json
```diff
@@ -37,6 +37,11 @@
     "eslint-config-airbnb-base": "^11.1.2",
     "eslint-plugin-import": "^2.2.0",
     "fs-extra": "^0.30.0",
+    "karma": "^1.5.0",
+    "karma-firefox-launcher": "^1.0.1",
+    "karma-mocha": "^1.3.0",
+    "karma-phantomjs-launcher": "^1.0.4",
+    "mocha": "^3.2.0",
```
WidgetsBurritos: Should we setup some sort of `npm run test` or something for this?

karma.conf.js
```diff
@@ -0,0 +1,69 @@
...
+
+
+    // frameworks to use
+    // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
+    frameworks: ['mocha'],
+
+
+    // list of files / patterns to load in the browser
+    files: [
+      'src/**/*.spec.js',
```
WidgetsBurritos: As far as locations for these tests, do we want to house the test for each pattern inside the same directory like all the javascript itself, or do we want a designated `test` directory for this?  

So for example, this file exists: 
`styleguide/derek/global-components/mainnav/mainnav.js`
Would we just want the spec right next to it?
`styleguide/derek/global-components/mainnav/mainnav.spec.js`

package.json
xinbin: aye definitely can. if anyone objects or has comments I'll edit it

package.json
```diff
@@ -33,15 +33,24 @@
...
     "browser-sync": "^2.18.8",
+    "chai": "^3.5.0",
     "eslint": "^3.18.0",
     "eslint-config-airbnb-base": "^11.1.2",
     "eslint-plugin-import": "^2.2.0",
     "fs-extra": "^0.30.0",
+    "karma": "^1.5.0",
+    "karma-firefox-launcher": "^1.0.1",
+    "karma-mocha": "^1.3.0",
+    "karma-phantomjs-launcher": "^1.0.4",
```
bighappyface: We can drop the phantomjs launcher in favor of using FF.

karma.conf.js
bighappyface: Yes. The `spec` segment of the filename helps delineate.

bighappyface: Shouldn't this path be `styleguide/**/*.spec.js` though?

package.json
xinbin: Sure we can drop phantom, I just thought we wanted to go with the headless approach.

karma.conf.js
xinbin: As per the path, I wasn't sure where we wanted tests to live, and I defaulted to separate space, much like the set up in RJL . I'll make this path change and remove phantom.

package.json
```diff
@@ -20,6 +20,9 @@
     "start": "concurrently 'npm run watch:babel' 'npm run start:harp' 'npm run start:bs'",
     "start:harp": "harp server styleguide",
     "start:bs": "browser-sync start --config bs-config.js",
+    "test": "npm run test:karma",
```
xinbin: Added these to match up with what exists in RJL

styleguide/src/demo.spec.js
```diff
@@ -0,0 +1,5 @@
+describe("A test suite", function() {
```
xinbin: This can go away when we add a real test.

package.json
bighappyface: We have headless FF in our jenkins environment :)

PR #428, opened 2017-04-04
RSWEB-8845: feat(zoolander): modularized features, lists & key offers

runrog:
Modularized 
product features
lists 
key offers

Seen here
http://localhost:3000/derek/solutions/solutions-usage


runrog:
@amberRrucker good catch, yes

runrog:
After talking w/Amber. We are leaving the typography page as is for now. Deeper issue there with layout inception.

PR #436, opened 2017-04-07
docs(global):cleanup docs

amberRrucker:
RSWEB-8784

Cleaning up some incorrect documentation.
Creating a separate task for the icons since it's going to require some work and hopefully we can move over the assets for that at that time.


brantdarilek:
👍

PR #437, opened 2017-04-12
feat(derek): Add leadership views

amberRrucker:
RSWEB-8768

http://localhost:3000/derek/incubation/
Viewable by clicking "Leadership Overview" Link

Adding views for leadership overview and interior page
Overview page needed images, name and title and a link to the interior page
Interior page needed image, name, title, link back to overview page and bio

Made adjustments to text partial and image partial in order to be able to support the view. They are minimal and won't effect www.

### DIFF
--- a/styleguide/derek/pattern-components/images/image.ejs
+++ b/styleguide/derek/pattern-components/images/image.ejs
@@ -1,0 +1,1 @@
+<img src="http://placekitten.com/300/400">
--- a/styleguide/derek/pattern-components/leads/lead-image/lead-image.ejs
+++ b/styleguide/derek/pattern-components/leads/lead-image/lead-image.ejs
@@ -1,0 +1,18 @@
+<div class="container standard-padding-top">
+  <div class="row flex-row">
+    <div class="col-md-9 flex-all">
+      <div class="subpanel">
+        <div class="subpanel-inner">
+          <%- partial("../lead/lead.ejs") %>
+        </div>
+      </div>
+    </div>
+    <div class="col-md-3 flex-all">
+      <div class="subpanel">
+        <div class="subpanel-inner">
+          <%- partial("../../images/image.ejs") %>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
--- a/styleguide/derek/view-templates/leadership/_leadership-individual.ejs
+++ b/styleguide/derek/view-templates/leadership/_leadership-individual.ejs
@@ -1,0 +1,7 @@
+<div class="leadership-individual">
+  <a href="/derek/view-templates/leadership/leadership-interior" class="leadership-attrWrapper">
+    <img src="http://placekitten.com/300/400" class="leadership-image">
+    <h4 class="leadership-name leadership-nameCenter">Mr. Mittens</h4>
+    <h6 class="leadership-title leadership-titleCenter">CEO</h6>
+  </a>
+</div>
--- a/styleguide/derek/view-templates/leadership/_leadership.scss
+++ b/styleguide/derek/view-templates/leadership/_leadership.scss
@@ -1,0 +1,51 @@
+.leadership-individual {
+  margin: 20px auto;
+}
+
+.leadership-image {
+  @include center-block;
+}
+
+.leadership-imageWrapper {
+  display: none;
+}
+
+.leadership-attrWrapper {
+  &:hover {
+    text-decoration: none;
+  }
+}
+
+.leadership-name {
+  margin-bottom: 0;
+}
+
+.leadership-nameCenter {
+  text-align: center;
+}
+
+.leadership-title {
+  margin: 0;
+}
+
+.leadership-titleCenter {
+  text-align: center;
+}
+
+.leadership-linkWrapper {
+  margin-top: 0;
+}
+
+.leadership-link {
+  @include standard-link;
+}
+
+@media only screen and (min-width: $screen-md-min) {
+  .leadership-imageWrapper {
+    display: block;
+  }
+
+  .leadership-linkWrapper {
+    margin-top: 20px;
+  }
+}
--- a/styleguide/_themes/derek/scss/_main.scss
+++ b/styleguide/_themes/derek/scss/_main.scss
@@ -31,6 +31,7 @@
 @import '../../../derek/pattern-components/ctas/cta';
 @import '../../../derek/view-templates/events/events';
 @import '../../../derek/view-templates/stories/stories';
+@import '../../../derek/view-templates/leadership/leadership';
 @import 'snowflakes/homepage';
 @import 'snowflakes/swatches';
 @import 'snowflakes/o365';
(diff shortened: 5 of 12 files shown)


styleguide/derek/pattern-components/leads/lead-image/lead-image.ejs
```diff
@@ -0,0 +1,18 @@
...
+    </div>
+    <div class="col-md-3 flex-all">
+      <div class="subpanel">
+        <div class="subpanel-inner">
+          <%- partial("../../images/image.ejs") %>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
```
amberRrucker: Added because it was needed and I forgot to during the hackathon. Same structure as lead-video.

styleguide/derek/view-templates/leadership/_leadership-individual.ejs
```diff
@@ -0,0 +1,9 @@
+<% if(!className) { var className = 'leadership-individual'; } %>
+
+<div class="<%- className -%>">
+  <a href="/derek/view-templates/leadership/leadership-interior" class="leadership-attrWrapper">
+    <img src="http://placekitten.com/200/150"/ class="leadership-image">
```
WidgetsBurritos: There's a weird slash after the src=""
```
src="http://placekitten.com/200/150"/
```

amberRrucker: Fixed!

styleguide/derek/pattern-components/images/image.ejs
```diff
@@ -0,0 +1 @@
+<img src="http://placekitten.com/300/400"/>
```
WidgetsBurritos: You don't need this trailing slash here. 

Also, just to confirm, the image on the overview and the image on individual leader pages are completely different aspect ratios.  Are there going to be two distinct images per leader on the website?

styleguide/derek/pattern-components/leads/lead-image/lead-image.ejs
WidgetsBurritos: This is fine

styleguide/derek/view-templates/leadership/_leadership.scss
```diff
@@ -0,0 +1,49 @@
...
+
+@media only screen and (min-width: $screen-xs-max) {
+  .leadership-individual {
+    &.center-block {
+      float: none;
+    }
+  }
+}
+
+@media only screen and (max-width: $screen-sm-max) {
```
WidgetsBurritos: Can we change this to 
```
@media only screen and (min-width: $screen-md-min) {
```

And then setup then set the below behavior as the default, and then override it for `md+` sizes. Eventually we're gonna start addressing media queries in RSWEB-8664, but the fewer queries we have to override at that point the better.

amberRrucker:
@WidgetsBurritos Fixed that feedback. Let me know if you see anything else.

styleguide/derek/pattern-components/images/image.ejs
amberRrucker: They didnt have the right aspect ration available in smaller size but Ill try to get closer.

amberRrucker: I just used the same mr.mittens and let the columns resize.

styleguide/derek/view-templates/leadership/_leadership.scss
amberRrucker: I fixed that one. The one that handles the lone-wolf leaders though won't override the floats properly when I try to do mobile-first. Bootstrap says no.

styleguide/derek/view-templates/leadership/_leadership.scss
```diff
@@ -0,0 +1,53 @@
...
+}
+
+.leadership-link {
+  @include standard-link;
+}
+
+@media only screen and (min-width: $screen-xs-max) {
+  .leadership-individual {
+    &.center-block {
+      float: none;
```
WidgetsBurritos: Since you're applying the center-block mixin to leadership-individual above it seems like we can get rid of all references to the `.center-block` class and that gets rid of any bootstrap related issues.

amberRrucker: It doesnt work. Ive run into it with some other patterns and if I use the mixin for columns then the mixin for center-block doesnt override properly. I can show you if you want to check it out.

WidgetsBurritos: Try this, and then swap out your `center-block` for `leadership-individualCenter`
```
.leadership-individualCenter {
  float: none;
  margin-left: auto;
  margin-right: auto;
}
```

styleguide/derek/view-templates/leadership/_leadership-individual.ejs
```diff
@@ -0,0 +1,9 @@
+<% if(!className) { var className = 'leadership-individual'; } %>
+
+<div class="<%- className -%>">
+  <a href="/derek/view-templates/leadership/leadership-interior" class="leadership-attrWrapper">
+    <img src="http://placekitten.com/300/400" class="leadership-image">
+    <h4 class="leadership-name center">Mr. Mittens</h4>
+    <h6 class="leadership-title center">CEO</h6>
```
WidgetsBurritos: We shouldn't use `center` here.  It should just be part of the class.

amberRrucker: I'm re-using the class on the interior page as well and the only difference between them is that this one is centered.

styleguide/derek/view-templates/leadership/_leadership.scss
amberRrucker: @WidgetsBurritos My question here is are we really going to use this? How are we actually doing to be adding these to the page in drupal? I imagine they would be panels that go into layouts? I think talking through that may help solve this issue.

styleguide/derek/view-templates/leadership/_leadership-individual.ejs
WidgetsBurritos: Well then maybe we setup `leadership-titleCenter` for maintainable purposes since this is inside the pattern itself.

styleguide/derek/view-templates/leadership/_leadership.scss
WidgetsBurritos: Well that brings up a good point.  There should be no layout mandated within the pattern itself.  So `leadership-individual` shouldn't set a width at all (meaning it wouldn't @include anything). 

Then inside each row in the `leadership.ejs`, we need to setup two new layouts that would also have to get created as part of the www ticket. We'd need the following layouts:
- Mobile 2-up that expands to a 4-up on desktop/table
- Mobile centered 1-up that is 50% wide, that goes to 25% on table/desktop. 

Layouts are fine to use core bootstrap classes as we already have that precedent set.

amberRrucker:
@WidgetsBurritos I made the adjustments we talked about in order to account for how we're going to implement in the CMS. That should cover everything.

WidgetsBurritos:
[run ci]

PR #438, opened 2017-04-14
RSWEB-8999: [Bug] Ceiling link font size inconsistent

xinbin:
Change summary:

* Set font size to a unified 1.15rem for ceiling

specifically, I put the vars in play, updated the `$navigation-font-size-medium` value so that it matches what the other nav items are. In line 86 I set that class to use the same var so that they're all unified. . . .

### DIFF
--- a/styleguide/derek/global-components/ceiling/_ceiling.scss
+++ b/styleguide/derek/global-components/ceiling/_ceiling.scss
@@ -98,7 +98,7 @@ $ceiling-font-size: $navigation-font-size-medium;
 .ceiling-menuInline-link {
   color: $white;
   display: block;
-  font-size: $navigation-font-size-medium;
+  font-size: 1.15rem;
   padding: 15px floor($ceiling-margin + 5);
   text-decoration: none;
   vertical-align: middle;
@@ -160,7 +160,7 @@ $ceiling-font-size: $navigation-font-size-medium;
 .ceiling-dropdownToggle {
   color: $white;
   display: block;
-  font-size: $navigation-font-size-medium;
+  font-size: 1.15rem;
   padding: 15px floor($ceiling-margin + 5);
   text-decoration: none;
 


WidgetsBurritos:
That variable doesn't just affect one spot.  So this will need to be confirmed in every spot: 

---

**Large Viewport** (looks as expected)
screen shot 2017-04-17 at 10 21 36 am
screen shot 2017-04-17 at 10 21 43 am

--

**Medium Viewport** (also changes navigation font sizes and sub navigation font size)
screen shot 2017-04-17 at 10 21 51 am
screen shot 2017-04-17 at 10 21 54 am

**Mobile view port** (also changes login/signup font sizes and "Request Free Quote" font size)
screen shot 2017-04-17 at 10 22 09 am
screen shot 2017-04-17 at 10 22 11 am

---

These don't bother me one way or another, but we need to make sure these are okay before we proceed. 
@amberRrucker ?

WidgetsBurritos:
[run ci]

styleguide/derek/global-components/ceiling/_ceiling.scss
```diff
@@ -160,7 +160,7 @@ $ceiling-font-size: $navigation-font-size-medium;
 .ceiling-dropdownToggle {
   color: $white;
   display: block;
-  font-size: $navigation-font-size-medium;
+  font-size: 1.15rem;
```
WidgetsBurritos: This looks like it's only changing the signup button, not the login button.  

We probably need to make this change to `.ceiling-menuInline-link` as well.

xinbin:
[run ci]

styleguide/derek/global-components/ceiling/_ceiling.scss
xinbin: 👍 done. rebuilt. loaded.

xinbin:
no_side-effects

PR #441, opened 2017-04-19
RSWEB-8978: pricing tables header bg color

writes:


writes:
I believe this has already been taken care of. Closing for now.

PR #443, opened 2017-04-21
bug(derek): fix LP footer issues

amberRrucker:
Visible on:
http://localhost:3000/derek/templates/landing-page/cta-LP

Can ensure it doesnt cause styling issues on pages with the expanded footer here:
http://localhost:3000/derek/templates/service/service-overview

- Remove region selector for landing pages
- Add bottom border to back to top so there is a division on pages without the expanded footer

### DIFF
--- a/styleguide/derek/global-components/back-to-top/_back-to-top.scss
+++ b/styleguide/derek/global-components/back-to-top/_back-to-top.scss
@@ -1,5 +1,6 @@
 .backtotop-wrapper {
   background-color: $gray-darkest;
+  border-bottom: 1px solid $gray-darker;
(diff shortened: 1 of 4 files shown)


WidgetsBurritos:
image

Is copyright info supposed to be right aligned on mobile? That looks off.

WidgetsBurritos:
image

^ Also it looks like any page with a pricing selector is broken.

amberRrucker:
@WidgetsBurritos Fixed those issues but I had to create a max width media query. I know we're trying to move away from that but it wasn't overriding the text-alignment so...

PR #444, opened 2017-04-25
refactor(derek): add class for padded container in zoolander

amberRrucker:
RSWEB-9033

Add a maintainable class for a padded container in zoolander. 
Switched out some of the containers with padding in zoolander to show the usage. 
However, because we havent gotten all the padding situation figured out with the standard-padding-full things look a little wonk on the solutions pages.

But the actual padded container class is on some of the patterns and is in modifiers.

### DIFF
--- a/styleguide/_themes/global/scss/modifiers.scss
+++ b/styleguide/_themes/global/scss/modifiers.scss
@@ -17,6 +17,11 @@ $uppercase: uppercase;
 img { max-width: 100%; }
 
 // Layout
+.derek-paddedContainer {
+  @include make-container;
+  padding: $standard-padding;
+}
+
 .full-width {
   width: $full-width;
 }
--- a/styleguide/derek/incubation/homepage.ejs
+++ b/styleguide/derek/incubation/homepage.ejs
@@ -1,6 +1,6 @@
 <%- partial("../pattern-components/headers/header-large-imac.ejs") %>
 
-<div class="container half-padding hostingOptions">
+<div class="derek-paddedContainer hostingOptions">
 	<div class="row half-padding-full">
 		<div class="col-md-11 center-block">
 			<h1 class="text-center no-margin no-padding" id="homepage-lead-top">Rackspace is the largest managed cloud provider</h1>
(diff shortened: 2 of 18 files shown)


brantdarilek:
@amberRrucker 
Is this correct?? Looks like an over abundance of padding here on the lists.
screen shot 2017-04-26 at 10 47 59 am

amberRrucker:
@brantdarilek its just in the documentation. Lists on their own will get that much padding but lists in mini panels. i.e. if they are right below features like in this example wont have that much padding. Just trying to make our containers and rivers have the same padding. If its in a mini panel they wont get as much. I hope that makes sense.

brantdarilek:
So like on the blog feed. It seem inconstant, no?
screen shot 2017-04-26 at 10 51 26 am

brantdarilek:
Also, looks like there might be an issue with resource cards.
screen shot 2017-04-26 at 10 56 12 am

amberRrucker:
@brantdarilek I added the new class to the classes that will be in boostrap containers. I didnt make a class for rivers or patterns that have a background color and padding and then a container inside them. Although the blog does get the same padding as containers with this new maintainable class. Its all just standard-padding. I guess we could add a class that would just be standard-padding? but it would be the same thing as applying the standard-padding class like it currently is.

We dont have a problem with the resource pattern. At some point it was decided that the standard-padding-full class wouldnt actually be full width anymore and since its applied to the templates in zoolander it pushes the solutions page in, causing problems with the layout there. It's an ongoing discussion but it has nothing to do with this PR.

amberRrucker:
Stinemetze and I talked about this in wheaties last week. Maybe we should hold on approving it or reviewing it until he comes back?

brantdarilek:
@amberRrucker 
Its up to you. Just reviewing this and had some questions what I was seeing since it was the first time.

amberRrucker:
@brantdarilek no it's cool. you bring up a good point. Something I wanna noodle on.

WidgetsBurritos:
Closing the PR for now, but will leave the branch intact. When this gets unblocked, we can just reopen the PR then.

PR #448, opened 2017-05-02
RSWEB-9041 RSWEB-9042 RSWEB-9043 RSWEB-9045 RSWEB-9046: Resource center

brantdarilek:
**This PR requires RSWEB-9084!**
This PR covers RSWEB-9041 RSWEB-9042 RSWEB-9043 RSWEB-9045 RSWEB-9046. 

The best way to review the design is by reviewing each of the pages here:
http://rsweb-9041-9042-9043-9045-9046-resource-center.styleguide.dev.website.rackspace.com/derek/incubation/


fonts/config.json
```diff
@@ -3049,6 +3049,76 @@
...
+      "svg": {
+        "path": "M633 500L433 667V333L633 500ZM1000 0V1000H0V0H1000ZM733 200H800V67H733V200ZM600 200H667V67H600V200ZM467 200H534V67H467V200ZM333 200H400V67H333V200ZM200 200H267V67H200V200ZM67 200H134V67H67V200ZM133 800H67V933H134V800ZM267 800H200V933H267V800ZM400 800H333V933H400V800ZM533 800H466V933H533V800ZM667 800H600V933H667V800ZM800 800H733V933H800V800ZM933 800H866V933H933V800ZM933 267H67V734H934V267ZM933 67H866V200H933V67Z",
+        "width": 1000
+      },
+      "search": [
+        "playbtn-film"
+      ]
+    },
+    {
+      "uid": "3b9dbe62ce1ded33cb5d6660c681bb08",
```
runrog: @brantdarilek I think you can remove this broken icon

brantdarilek: Done.

WidgetsBurritos:
screen shot 2017-05-03 at 11 04 09 pm
screen shot 2017-05-03 at 11 04 15 pm

Some of the break points have overlap issues.  It almost looks intentional on some of the resources, but on videos specifically it could be a bit problematic.

styleguide/_layouts/default.jade
```diff
@@ -18,7 +18,7 @@ html(lang="en")
     link(href="#{ fonts }", rel="stylesheet")
     link(href="css/main.css", rel="stylesheet")
     link(href="https://fonts.googleapis.com/css?family=Open+Sans:400,600", rel="stylesheet")
-    link(href="https://maxcdn.bootstrapcdn.com/font-awesome/4.6.1/css/font-awesome.min.css" rel="stylesheet")
+    link(href="https://fonts.googleapis.com/css?family=Khand" rel="stylesheet")
```
WidgetsBurritos: Is it just me, or does Khand looks strange now?

Before: 
screen shot 2017-05-03 at 11 29 17 pm

After:
screen shot 2017-05-03 at 11 29 13 pm

Maybe google just renders it differently or something, but this is definitely a big difference here that I want to make sure we're aware of, as it may create discrepancies between www and zoolander.

xinbin: The first is definitely bold or something. It's thicker.

brantdarilek:
@WidgetsBurritos 
I've updated to accommodate for the tablet breakpoint. (See below images)
As for khand, I can't speak for the differences. I do think its better to include it than not. Not including it seems like more of a risk for discrepancies. 
@amberRrucker @runrog 
Do we host our own version on a CDN? If so I would just pull in that version instead of from google.

screen shot 2017-05-04 at 10 17 07 am
screen shot 2017-05-04 at 10 17 28 am

runrog:
@brantdarilek don't think we have our own CDN link but in zoolander, try importing the rest of the Khand weights like this
```
<link href="https://fonts.googleapis.com/css?family=Khand:300,400,500,600,700" rel="stylesheet">
```
Looks like you are only including the standard weight

amberRrucker:
@brantdarilek We used to have it included but it was pulled out to avoid double loading the font into www. @WidgetsBurritos I'm preeeeetty sure that was your request but I dont remember.

WidgetsBurritos:
Looking into it a bit more, the discrepancy is between my local version of Khand and the version provided by google fonts.  I've disabled my local version of khand for now.  

I'd say keep the khand reference in here, but maybe make sure you have the necessary weights per Roger's comment, although I'm not sure we need all of those weights.  I honestly can't remember.  

Given that this will only affect zoolander (as www includes the font differently), I'm okay with it being a little less performant, as long as it provides consistency with what's in www. 

Otherwise I'm generally fine with this but do have one question.  I understand hiding the resource images on mobile, but does that make sense for the video resource?   There's no way to actually get to the video otherwise. 

image

brantdarilek:
@WidgetsBurritos @runrog @amberRrucker 
I've updated the font to include all weights.
I've also updated the template so not hide the video on mobile.

WidgetsBurritos:
@brantdarilek 

It's almost there.  It appears we still have a content overlap on from 992px to 1279px:

image

brantdarilek:
@WidgetsBurritos 
Should be good now.

writes:
+1

WidgetsBurritos:
👍  so we can work on the www without much conflict.

PR #445, opened 2017-04-25
refactor(derek): revamp resources pattern

amberRrucker:
RSWEB-8629

Refactoring and revamping the resources pattern.
I've made the pattern maintainable and chosen maintainable classes for the icons to avoid conflicts.
The EBPs should still just be able to choose the same classes, we can just fill them into the middle of the class.

This is viewable on any page that has resources on it but I used this one:
http://localhost:3000/derek/templates/product/product-overview

Please review in mobile and with the hover state as well.

PR #452, opened 2017-05-09
fix(resource-center): Adjust icons colors

WidgetsBurritos:
This PR is a half way step between #445 which was already merged and #448 which has yet to be merged.  This is needed to correct some class related issues that happened in between these two steps, but are needed to be resolved independently of that other work.


WidgetsBurritos:
@amberRrucker He was the one who gave me the list.

amberRrucker:
@WidgetsBurritos perf

PR #459, opened 2017-05-16
RSWEB-8951 feat(newpattern): addition of service level teaser

runrog:
Addition of Service level teaser pattern.
Seen here: 
http://localhost:3000/derek/solutions/solutions

### DIFF
--- a/styleguide/derek/pattern-components/service-level-teaser/_service-level-teaser.scss
+++ b/styleguide/derek/pattern-components/service-level-teaser/_service-level-teaser.scss
@@ -1,0 +1,76 @@
+.serviceLevelTeaser {
+  background: $white;
+  border-bottom: 4px solid $brand-primary;
+  border-top: 4px solid $brand-primary;
+  box-shadow: 0 1px 3px $black-shadow;
+  display: block;
+  margin-bottom: 10px;
+  padding: 25px;
+  position: relative;
+  text-decoration: none;
+  width: 100%;
+
+  &:hover {
+    background-color: $gray-lightest;
+    text-decoration: none;
+  }
+}
+
+.serviceLevelTeaser-body {
+  padding-bottom: 10px;
+}
+
+.serviceLevelTeaser-col-full {
+  @include make-md-column(12);
+}
+
+.serviceLevelTeaser-header {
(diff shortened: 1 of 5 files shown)


WidgetsBurritos:
[run ci]

styleguide/derek/pattern-components/service-level-teaser/_service-level-teaser.scss
```diff
@@ -0,0 +1,80 @@
...
+  width: 100%;
+
+  &:hover {
+    background-color: $gray-lightest;
+    text-decoration: none;
+  }
+
+  a {
+    @include standard-link;
+  }
```
WidgetsBurritos: Insert typical comment about, is this nested anchor absolutely necessary here? Shouldn't we just be targetting `.serviceLevelTeaser-bottomLink`, or am I misunderstanding the use case?

styleguide/derek/pattern-components/service-level-teaser/service-level-teaser.ejs
```diff
@@ -0,0 +1,102 @@
...
+  </div>
+</div>
+
+<div class="half-padding container">
+  <h3>Service Level Teasers 2 only (default)</h3>
+</div>
+
+<!-- Teaser with only 2 option -->
+<div class="half-padding">
+  <div class="container">
```
WidgetsBurritos: It's worth noting that either this PR, or #444 will need to be adjusted (depending on merge order), to allow for usage of the new `.derek-paddedContainer` class. It doesn't require immediate action, but we'll need to make sure we remember that come merge time.

styleguide/derek/pattern-components/service-level-teaser/_service-level-teaser.scss
runrog: @WidgetsBurritos oh nice catch. This is not needed.

styleguide/derek/pattern-components/service-level-teaser/service-level-teaser.ejs
runrog: Ok noted.

PR #462, opened 2017-05-22
feat(derek): add dark ceiling

amberRrucker:
RSWEB-9183

Visible here:
http://localhost:3000/derek/templates/login

Looking for dark ceiling matching the screen shot in the ticket. The page content doesn't match the screen shot as it is not a requirement.


styleguide/derek/global-components/ceiling/ceilingDark/_ceilingDark.scss
```diff
@@ -0,0 +1,69 @@
...
+  .navbar-chaticonDark,
+  .navbar-emailiconDark {
+    padding: 15px 0;
+  }
+}
+
+@media only screen and (max-width: $screen-sm-max) {
+  .basement-iconListItem {
+    font-size: 1rem;
+  }
```
WidgetsBurritos: Why are there basement changes in this file?

styleguide/derek/global-components/ceiling/ceilingDark/_ceilingDark.scss
```diff
@@ -0,0 +1,69 @@
...
+}
+
+@media only screen and (min-width: $screen-md-min) {
+  margin: 0 (-$grid-gutter-width / 2);
+}
```
WidgetsBurritos: What is this applying a margin to? It's not inside a selector so it doesn't seem like it's even doing anything here.

styleguide/derek/global-components/ceiling/ceilingDark/_ceilingDark.scss
```diff
@@ -0,0 +1,69 @@
...
+  background-size: 100% 79%;
+  float: left;
+  height: 35px;
+  margin: 5px 10px;
+  position: relative;
+  transition-duration: .3s;
+  width: 150px;
+}
+
+@media only screen and (max-width: $screen-xs-max) {
```
WidgetsBurritos: Can you reverse this to use min-width instead of max-width?

styleguide/derek/global-components/ceiling/ceilingDark/_ceilingDark.scss
```diff
@@ -0,0 +1,69 @@
+.ceiling-menuInline-item {
+  border-left: 1px solid $white;
+}
```
WidgetsBurritos: This should use a selector specific to the dark ceiling.  It's coloring the border for the standard landing pages. 

image

amberRrucker:
@WidgetsBurritos, since we aren't really styling the actual login page from this PR I didn't want to list it as an example page. No EBPs or other users should be using that theme so I don't really want it to be in the documentation.

amberRrucker:
@WidgetsBurritos I made those changes and made the same mobile first change to the landing page ceiling.

amberRrucker:
[run ci]

PR #470, opened 2017-06-07
feat(karma): pivot to phantomjs and run karma via travis

bighappyface:
This PR pivots `karma` to use PhantomJS to avoid complications surrounding headless Firefox with TravisCI.


bighappyface:
@xinbin @WidgetsBurritos this small PR is ready for review.

The Travis run of `test:karma` is failing correctly as the only test in the code base is setup to fail:

https://github.com/rackerlabs/zoolander/blob/997fe49254c4b0cc075c38522424b2f3d6cba2ef/styleguide/src/demo.spec.js#L4

PR #471, opened 2017-06-12
Filter Unit Testing

xinbin:
Description: Add tests for filter js


WidgetsBurritos:
@xinbin Closing this PR since we're going with another approach.  Not going to delete the branch though in case we need it for reference.

PR #473, opened 2017-06-13
RSWEB-9122: Add node 6 to Travis

WidgetsBurritos:
This PR tests adding node 6 into the travis mix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 node_js:
   - "4.2.4"
+  - "6.10.0"
 env:
   - CMD="sass:compile"
   - CMD="lint:sass"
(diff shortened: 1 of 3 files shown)


WidgetsBurritos:
image

WidgetsBurritos:
image

This build includes updated node 6 version and includes shrinkwrap file. 

^ Build: https://travis-ci.org/rackerlabs/zoolander/builds/242570132

WidgetsBurritos:
It's worth noting, I have seem some general randomness around travis itself of about 1 or so minutes.  That may be related to availability of resources, but I've consistently seen faster times in Node 6 vs Node 4 and also improvements on both by using the shrinkwrap file.

Another interesting note, is that I haven't had to restart any of the node 6 tests because of `ECONNRESET` problems, but I have had do that several times with the node 4 tests.  That may have been sheer luck, but also seems indicative of a positive side effect of reducing the amount of connections needed.

PR #474, opened 2017-06-20
Filterjs unit tests

xinbin:
Break up filterJs a bit to allow for unit testing.
Add unit tests.

Based tests on what the values come out to when you change the size of the window and whether youre scrolled down or not.

### DIFF
--- a/styleguide/derek/global-components/filter-bar/filter-bar.js
+++ b/styleguide/derek/global-components/filter-bar/filter-bar.js
@@ -1,51 +1,93 @@
+/* eslint no-use-before-define: 0 */
+
+const Zoolander = Zoolander || {};
+
+Zoolander.rsFilterBar = {
+  calcAtBottom: (dimensions) => {
+    const defaults = {
+      filterBarOffsetTop: 0,
+      filterBarOuterHeight: 0,
+      filterFormOuterHeight: 0,
+      mainNavHeight: 0,
+      windowTop: 0,
+    };
+    const values = Object.assign(defaults, dimensions);
+    const filterFormHeight = values.filterFormOuterHeight;
+    const filterBottom = values.filterBarOuterHeight +
+                         values.filterBarOffsetTop;
+    const formBottom = values.windowTop + filterFormHeight + values.mainNavHeight + 10;
+    return formBottom >= filterBottom;
+  },
+  calcInBounds: (dimensions) => {
+    const defaults = {
+      filterBarOffsetTop: 0,
+      filterBarOuterHeight: 0,
+      filterFormOffsetTop: 0,
+      filterFormOuterHeight: 0,
+      isInBounds: false,
+      mainNavHeight: 0,
+      stickyItemTop: 0,
+      windowTop: 0,
+    };
+    const values = Object.assign(defaults, dimensions);
+    // wt = pixels from top of window to top of scrolled area
+    const filterFormHeight = values.filterFormOuterHeight;
+    if (!values.isInBounds) {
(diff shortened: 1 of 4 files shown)


styleguide/derek/global-components/filter-bar/filter-bar.js
```diff
@@ -1,13 +1,63 @@
...
+      filterFormOffsetTop: 0,
+      filterFormOuterHeight: 0,
+      isInBounds: false,
+      mainNavHeight: 0,
+      stickyItemTop: 0,
+      windowTop: 0,
+    };
+    const values = Object.assign(defaults, dimensions);
+    // wt = pixels from top of window to top of scrolled area
+    const filterFormHeight = parseInt(values.filterFormOuterHeight, 10);
```
bighappyface: @xinbin do tests still work with the `parseInt` calls removed? I suspect they are not necessary.

styleguide/derek/global-components/filter-bar/filter-bar.js
```diff
@@ -1,13 +1,63 @@
...
+  calcAtBottom: (dimensions) => {
+    const defaults = {
+      filterBarOffsetTop: 0,
+      filterBarOuterHeight: 0,
+      filterFormOuterHeight: 0,
+      mainNavHeight: 0,
+      windowTop: 0,
+    };
+    const values = Object.assign(defaults, dimensions);
+    const filterFormHeight = parseInt(values.filterFormOuterHeight, 10);
```
bighappyface: @xinbin do tests still work with the `parseInt` calls removed? I suspect they are not necessary.

WidgetsBurritos: I believe at some point we had something like `363px` getting passed in and this was meant to strip the units off. But looking at the jQuery results for both the `$.offset()` and `$.outerHeight()` functions, which should be populating this data, it's probably unnecessary. I'm fine with removing it, we just need to test it in multiple browsers (i.e. IE -- Ayeeeee!) to make sure everything still works as expected.

styleguide/derek/global-components/filter-bar/filter-bar.js
```diff
@@ -17,35 +67,27 @@
 
     // Callback to check if inbounds of side menu
     function checkInBounds(wt) {
```
WidgetsBurritos: Can you do me one more favor on this function? 

The only time window top is calculated is right before the function is called.  Meaning, there's zero difference in passing that value in, versus just calculating it inside the function.  
```
          const $rsWindowTop = parseInt($rsWindow.scrollTop(), 10);
          inBounds = checkInBounds($rsWindowTop);
```

Can we just get rid of the parameter, and then change the array item below to this?:

```
windowTop: $rsWindow.scrollTop(),
```

WidgetsBurritos:
It appears something is slightly broken now.  If you make your screen height a little smaller and scroll up and down you get jittery behavior. 

Zoolander: 
zoolander

Production:
rackspace com

xinbin:
Not seeing this jitter issue, I rebased it to bring in any changes. Going to test some more on different browsers and compare against the dev/branch link

xinbin:
I take that back. I was playing with it and figured out how to replicate the issue. I wish I had commited each change separately .. now I can't step back to what part of the refactoring "broke" it.

WidgetsBurritos:
@xinbin let's find some time to crash this today.  we should be able to figure it out.

xinbin:
@runrog  please take a look at the code change + let me know if you have feedback

PR #475, opened 2017-06-20
RSWEB-9097: Add tracking to filtering

brantdarilek:
https://jira.rax.io/browse/RSWEB-9097

### DIFF
--- a/styleguide/derek/global-components/tracking/tracking.js
+++ b/styleguide/derek/global-components/tracking/tracking.js
@@ -181,6 +181,35 @@ Zoolander.Tracking = (function Tracking($) {
         eventNonInteraction: 0,
       });
     });
+
+    $('.rsFilter-form').on('submit', (e) => {
+      dataLayer.push({
+        event: 'ga.event',
+        eventCategory: 'Resource Center',
+        eventAction: 'Resource Center Filter - Product',
+        eventLabel: $(e.target).find('option:selected').text(),
+        eventValue: '0',
+        eventNonInteraction: 0,
+      });
+
+      dataLayer.push({
+        event: 'ga.event',
+        eventCategory: 'Resource Center',
+        eventAction: 'Resource Center Filter - Type',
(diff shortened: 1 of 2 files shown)


styleguide/derek/global-components/tracking/tracking.js
```diff
@@ -181,6 +181,35 @@ Zoolander.Tracking = (function Tracking($) {
...
+      dataLayer.push({
+        event: 'ga.event',
+        eventCategory: 'Resource Center',
+        eventAction: 'Resource Center Filter - Type',
+        eventLabel: getCheckboxes($(e.target).find('input:checked')),
+        eventValue: '0',
+        eventNonInteraction: 0,
+      });
+    });
+  }
```
WidgetsBurritos: I'm not seeing the event firing here: https://rax.io/rsweb-9097-preview

Is it working for you locally? Additionally, since we're starting to use zoolander to manage data tracking, does it make sense to include GTM inside zoolander itself? That would make it a bit easier to test this kind of stuff.

styleguide/derek/global-components/tracking/tracking.js
```diff
@@ -181,6 +181,35 @@ Zoolander.Tracking = (function Tracking($) {
...
+
+  function getCheckboxes(elements) {
```
WidgetsBurritos: Can we rename this to something a little bit more specific? Something like `getCheckboxValuesByElement()`.  It is a bit more verbose, but it better explains what the function is doing.

styleguide/derek/global-components/tracking/tracking.js
brantdarilek: Not sure why, but the my changes are not showing up in the www zoolander branch build.
Are there config changes that need to be made there, since moving over tracking?
screen shot 2017-06-27 at 1 47 51 pm

WidgetsBurritos:
[run ci]

PR #476, opened 2017-06-21
fix(bug): resource button style after visit

xinbin:
Fix resource visited button state..
Before:
current

After:
change-goes-back-towhat_it_was


brantdarilek:
+1 on 9ba5158

PR #479, opened 2017-07-03
fix(filter): bring back parseInt

WidgetsBurritos:
This PR adds `parseInt()` back when using `$.css()`, as it's needed: 

e.g.
```
$('body').css('width');
"1196px"
```


xinbin:
1+

PR #480, opened 2017-07-06
feat(derek): add new stories resource card

amberRrucker:
RSWEB-9360
Creating a new resources card for customer stories.

Viewable here:
http://localhost:3000/derek/view-templates/resource-center/resource-center

Should appear the same as the other resources cards aside from being signup green and with a new quote icon.


brantdarilek:
👍 on a1d682c

PR #481, opened 2017-07-10
RSWEB-9346: Search result styles

brantdarilek:
https://jira.rax.io/browse/RSWEB-9346

### DIFF
--- a/styleguide/derek/view-templates/search/search.ejs
+++ b/styleguide/derek/view-templates/search/search.ejs
@@ -1,0 +1,54 @@
+<!--Header Medium-->
+<div class="imacBanner">
+  <div class="imacBanner-container">
+    <div class="imacBanner-textRow">
+      <div class="imacBanner-textContainer imacBanner-headlineMedium">
+        <h2 class="imacBanner-headline">Search Results</h2>
+        <span class="imacBanner-line left-top"></span>
+        <span class="imacBanner-line left-bottom"></span>
+        <span class="imacBanner-line right-bottom"></span>
+      </div>
+    </div>
+  </div>
+</div>
+
+<div class="container half-padding">
+  <!-- Lead pattern -->
+  <div class="lead-wrapper">
+    <div class="lead-headerWrapper text-left">
+      <h1 class="lead-header">Search Results for: Cloud</h1>
+    </div>
+    <div class="lead-bodyWrapper text-left">
+      <p>About 6,030 results (0.31 seconds)</p>
+    </div>
+  </div>
+
+  <%- partial('_search-result') %>
+  <%- partial('_search-result') %>
+  <%- partial('_search-result') %>
+  <%- partial('_search-result') %>
+  <%- partial('_search-result') %>
+  <%- partial('_search-result') %>
+  <%- partial('_search-result') %>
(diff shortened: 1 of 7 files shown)


styleguide/derek/view-templates/search/search.ejs
```diff
@@ -0,0 +1,54 @@
...
+      <li class="search-pagerItem"><a href="javascript:void(0);">2</a></li>
+      <li class="search-pagerItem"><a href="javascript:void(0);">3</a></li>
+      <li class="search-pagerItem"><a href="javascript:void(0);">4</a></li>
+      <li class="search-pagerItem"><a href="javascript:void(0);">5</a></li>
+      <li class="search-pagerItem"><a href="javascript:void(0);">6</a></li>
+      <li class="search-pagerItem"><a href="javascript:void(0);">7</a></li>
+      <li class="search-pagerItem"><a href="javascript:void(0);">8</a></li>
+      <li class="search-pagerItem"><a href="javascript:void(0);">9</a></li>
+      <li class="search-pagerItem"><a href="javascript:void(0);">10</a></li>
+    </ul>
```
pobtastic: Needs a "search-pagerItem-current" or similar to show the current page.

pobtastic: Also, are we going to keep/ lose the next/ previous links?

pobtastic:
Please could we also add the search box itself.

styleguide/derek/view-templates/search/search.ejs
```diff
@@ -0,0 +1,54 @@
...
+      </div>
+    </div>
+  </div>
+</div>
+
+<div class="container half-padding">
+  <!-- Lead pattern -->
+  <div class="lead-wrapper">
+    <div class="lead-headerWrapper text-left">
+      <h1 class="lead-header">Search Results for: Cloud</h1>
```
pobtastic: Do we want to transform this to uppercase as we do with the current search results?  Or maybe wrap the search term in `<em>` (we don't do that, but it's fairly commonplace I think?)

styleguide/derek/view-templates/search/search.ejs
WidgetsBurritos: Actually, now that you bring up the pager...  I kinda wonder if it makes sense to just follow the lead from resource center, and use the default pager out of the box. 

image

I know we're all about the maintainable approach on things, but I think we also should consider consistency, which we won't get with this approach. 

Just my two cents.  I'm not married to it one way or another, but just thought I'd offer up that feedback.

brantdarilek: The pager is what I pulled from the current dev branch for the solr build.
http://rsweb-9347-solr.www.dev.website.rackspace.com/

If there is an alternate version, I think it should happen there so that I at least would know the capabilities.

styleguide/derek/view-templates/search/search.ejs
brantdarilek: @pobtastic  I showed this to the CROUX team yesterday and everyone gave a +1 as is, since its using our current styles. It was discussed adding more at a later date.

PR #482, opened 2017-07-11
RSWEB-9334 feat(new-pattern): added 3 featured customer stories to its homepage

runrog:
You can see it in action locally here
http://localhost:3000/derek/view-templates/stories/stories

### DIFF
--- a/styleguide/derek/global-components/anchor-scroll/anchor-scroll.js
+++ b/styleguide/derek/global-components/anchor-scroll/anchor-scroll.js
@@ -1,0 +1,28 @@
+// This plugin can be used to anchor scroll to anything. Usage:
+// 1) Give your link the "rsAnchor-scroll" class
+// 2) Give the href of your link a hashed id (ie href="#someID")
+// 3) Set that id to the element you want scrolled to (ie id="someID")
+(($) => {
+  $.fn.anchorScroll = function anchorScroll() {
+    const plugin = this;
+    $(this).click(function scrollTo(e) {
+      e.preventDefault();
+      // we need to subtract the top nav or else scrolled section will look cut off.
+      const $this = $(this.hash);
+      const $top = $this.offset().top;
+      const dimensions = {
+        nav: $('.navbar-fixed-top').outerHeight(),
+        top: $top,
+      };
+      if (typeof $this.offset() !== typeof undefined && $this.offset() !== false) {
+        $('html, body').animate({
+          scrollTop: plugin.anchorScroll.calcTopScroll(dimensions),
+        }, 500);
+      }
+    });
+    $.fn.anchorScroll.calcTopScroll = function calcTopScroll(d) {
+      return d.top - d.nav;
+    };
+    return this;
+  };
+})(jQuery);
--- a/styleguide/derek/view-templates/stories/_stories.scss
+++ b/styleguide/derek/view-templates/stories/_stories.scss
@@ -1,3 +1,110 @@
+// Customer Stories Feature Colors
+$default-bg-img: 'https://752f77aa107738c25d93-f083e9a6295a3f0714fa019ffdca65c3.ssl.cf1.rackcdn.com/bananers/IMACbanners/collaboration-2.jpg';
+// Mixin for featured stories bg types
+@mixin featuredBg($class, $gradient) {
+  #{$class} {
+    &::before {
+      background-image: url($default-bg-img);
+    }
+
+    &::after {
+      background-image: $gradient;
+    }
+  }
+}
+
+.story-featured-container {
+  background-color: $gray-lightest;
+  padding: 40px 25px;
+}
+
+.story-featured-text {
+  @include make-container;
+  text-align: center;
+}
+
+.story-featured-cards {
+  @include make-container;
+}
+
+.story-featured-cardRow {
+  @include make-row;
+  margin-top: 10px;
+}
+
+.story-featured-cardLogo {
+  @include make-md-column(3);
+  padding-left: 0;
+  padding-right: 0;
+}
+
+.story-featured-logoInner {
+  @include flex-center;
+  box-shadow: 0 1px 3px transparentize($black, .5);
+  display: flex;
+  height: 220px;
+}
+
+.story-featured-cardQuote {
+  display: none;
+}
+
+.story-featured-link {
+  &:hover {
+    &::after {
+      background-color: $gray-light;
+      opacity: 1;
+      transition-duration: .3s;
+    }
+  }
+}
+
+// 3 featured logo / quote hover states
+.story-featured-cardLogoTop {
+  background-color: darken($rackspace-red, 5%);
+}
+
+.story-featured-cardLogoCenter {
+  background-color: $teal-dark;
+}
+
+.story-featured-cardLogoBottom {
+  background-color: $purple;
+}
+
+.story-featured-seeMore {
+  color: transparentize($black, .5);
+  display: block;
+  font-weight: $font-weight-black;
+  margin-top: 50px;
+  text-align: center;
+
+  &::before {
+    content: 'See More Customer Stories';
+    display: block;
+    font-family: $heading;
+    font-size: 20px;
+  }
+
+  &::after {
+    content: '\e806';
+    display: block;
+    font-family: 'rswebfonts';
+    font-size: 20px;
+  }
+
+  &:hover {
+    color: transparentize($black, .3);
+    text-decoration: none;
+    transition-duration: .2s;
+  }
+
+  &:focus {
+    color: transparentize($black, .5);
+    text-decoration: none;
+  }
+}
+
 .story-container {
   @include make-container;
   padding: $standard-padding;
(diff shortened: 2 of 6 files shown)


styleguide/derek/view-templates/stories/_stories.scss
```diff
@@ -1,3 +1,218 @@
+// Customer Stories Feature Colors
+$cust-stories-red: #9d0000;
+$cust-stories-teal: #00b4c3;
+$cust-stories-purple: #563777;
+$default-bg-img: 'https://752f77aa107738c25d93-f083e9a6295a3f0714fa019ffdca65c3.ssl.cf1.rackcdn.com/bananers/IMACbanners/collaboration-2.jpg';
```
brantdarilek: Are we only using one background?

styleguide/derek/view-templates/stories/_stories.scss
```diff
@@ -1,3 +1,218 @@
+// Customer Stories Feature Colors
+$cust-stories-red: #9d0000;
+$cust-stories-teal: #00b4c3;
+$cust-stories-purple: #563777;
```
brantdarilek: Should we create new variables here, or add them to the swatches?

styleguide/derek/view-templates/stories/_stories.scss
```diff
@@ -1,3 +1,218 @@
...
+    &::after {
+      opacity: 1;
+      transition-duration: .3s;
+    }
+  }
+}
+
+// 3 featured logo / quote hover states
+.story-featured-cardTop {
+  .story-featured-cardLogo {
```
WidgetsBurritos: Instead of nesting here, can we just have something like this?

`.story-featured-cardTopLogo`

It's just a div, so it doesn't seem like we're trying to offset for bootstrap or anything here, so I'd prefer no nesting. 

The same applies to the other nested classes below.

styleguide/derek/view-templates/stories/_stories.scss
```diff
@@ -1,3 +1,218 @@
...
+    &:hover {
+      &::after {
+        background-image: linear-gradient(to bottom right, $cust-stories-purple, $cust-stories-purple);
+      }
+    }
+  }
+}
+
+// Desktop only rules
+@media (min-width: $screen-md-min) {
```
WidgetsBurritos: Thank you for the mobile-first approach here.  Would you mind moving this code to the bottom of the file with the rest of the media queries?

styleguide/derek/view-templates/stories/stories.ejs
```diff
@@ -4,19 +4,55 @@ var images = [
...
+  },
+  {
+    area: 'Bottom',
+    logo: 'https://www.rackspace.com/sites/default/files/story-logos/docker_0.png',
+  }
+]
 %>
 
 <!--Header Short-->
-<div class="imacBanner imacBanner-stories">
```
WidgetsBurritos: So just for my information, are we completely dropping imac banners on the stories page?

styleguide/derek/view-templates/stories/_stories.scss
runrog: So this is just the default if no image is selected. The user should have the option to override this by choosing the image in Drupal.

styleguide/derek/view-templates/stories/_stories.scss
runrog: I'm cool with either or, @amberRrucker you wanna weigh in on this one?

amberRrucker: We can't use the ones we have? Or lighten or darken the existing swatches? I'd hate to have these as hex values when they are so close to what we already have.

styleguide/derek/view-templates/stories/stories.ejs
runrog: Yes we are removing them completely on this page

amberRrucker: yessir

WidgetsBurritos: 👍

styleguide/derek/view-templates/stories/_stories.scss
runrog: @amberRrucker we could with the red and teal but not the purple, it doesn't exist

runrog:
Ok, new changes are up. I also forgot to add the functionality to the scroll to more stories anchor link. That has been added as well.

styleguide/derek/view-templates/stories/_stories.scss
```diff
@@ -1,3 +1,92 @@
...
+    content: 'See More Customer Stories';
+    display: block;
+    font-family: $heading;
+    font-size: 20px;
+  }
+
+  &::after {
+    content: '\f0d7';
+    display: block;
+    font-family: 'FontAwesome';
```
brantdarilek: @runrog 
You need to specify the `rswebfonts` here.
screen shot 2017-07-12 at 11 17 42 am
screen shot 2017-07-12 at 11 18 57 am

brantdarilek: Also, it looks like its picking up a `:visited` and remaining bold after being clicked.

runrog: fixed

styleguide/global/global.js
```diff
@@ -32,6 +32,28 @@ jQuery(document).ready(($) => {
...
+      const $this = $(this.hash);
+      if (typeof $this.offset() !== typeof undefined && $this.offset() !== false) {
+        $('html, body').animate({
+          scrollTop: ($this.offset().top - fixedNav),
+        }, 500);
+      }
+    });
+    return this;
+  };
+})(jQuery);
```
WidgetsBurritos: Plugins shouldn't live in global.js.  Specifically not page-specific plugins like this.  Since we're only using this inside customer stories, it really should live in it's own file.

styleguide/global/global.js
```diff
@@ -32,6 +32,28 @@ jQuery(document).ready(($) => {
...
+  $.fn.anchorScroll = function anchorScroll() {
+    $(this).click(function scrollTo(e) {
+      e.preventDefault();
+      // we need to subtract the top nav or else scrolled section will look cut off.
+      const fixedNav = $('.navbar-fixed-top').outerHeight();
+      const $this = $(this.hash);
+      if (typeof $this.offset() !== typeof undefined && $this.offset() !== false) {
+        $('html, body').animate({
+          scrollTop: ($this.offset().top - fixedNav),
+        }, 500);
```
WidgetsBurritos: I know this calculation is simple, but if we don't write this to be unit testable now, we set ourselves up for trouble down the road.  

I'd recommend creating a separate function called something like `calculateAnchorScrollTop()`, that takes, top and fixed nav as parameters.  

Then we need to write unit tests to tests what happens when various values are sent into this function. 

You can use filter-bar as a point of reference for writing the tests...
https://github.com/rackerlabs/zoolander/blob/master/styleguide/derek/global-components/filter-bar/filter-bar.spec.js

styleguide/global/global.js
runrog: I put this one in global because it's not page specific. It's a plugin we can use globally to anchor scroll to anything, not just in customer stories.

runrog: I can move this into a single file under global-components

WidgetsBurritos: We still need the option to attach it conditionally, so global-components is the place to put it.

runrog:
Unit test is now up and it is js is separated

styleguide/derek/view-templates/stories/_stories.scss
```diff
@@ -149,3 +244,122 @@
     max-width: 120px;
   }
 }
+
+// Desktop only rules
+@media (min-width: $screen-md-min) {
+  .story-freatured-cardRow {
```
amberRrucker: freatured??

styleguide/derek/view-templates/stories/_stories.scss
```diff
@@ -149,3 +244,122 @@
...
+
+  .story-featured-cardQuoteCenter {
+    &::before {
+      background-image: url($default-bg-img);
+    }
+
+    &::after {
+      background-image: linear-gradient(to bottom left, transparentize($teal-dark, .4), $teal-dark);
+    }
+  }
```
amberRrucker: Could we make a mixin for these?

runrog: done

brantdarilek:
+1 on df0eaf5

styleguide/derek/global-components/anchor-scroll/anchor-scroll.js
```diff
@@ -0,0 +1,30 @@
...
+      e.preventDefault();
+      // we need to subtract the top nav or else scrolled section will look cut off.
+      const $this = $(this.hash);
+      const $top = $this.offset().top;
+      const dimensions = {
+        nav: $('.navbar-fixed-top').outerHeight(),
+        el: $this,
+        top: $top,
+      };
+      plugin.anchorScroll.caclTopScroll(dimensions);
```
WidgetsBurritos: calc*

styleguide/derek/global-components/anchor-scroll/anchor-scroll.js
```diff
@@ -0,0 +1,30 @@
...
+    });
+    $.fn.anchorScroll.caclTopScroll = (d) => {
+      if (typeof d.el.offset() !== typeof undefined && d.el.offset() !== false) {
+        $('html, body').animate({
+          scrollTop: (d.top - d.nav),
+        }, 500);
```
WidgetsBurritos: Since this function is for "calculating" it should just do and return the calculated value.   So something like this: 

```
$.fn.anchorScroll.caclTopScroll = (top, nav) => {
  return top - nav;
}
```

Then just move the animate code to the area above in place of `plugin.anchorScroll.caclTopScroll(dimensions);`

```
$('html, body').animate({
  scrollTop: plugin.anchorScroll.calcTopScroll(d.top, d.nav),
}, 500);
```

runrog: @WidgetsBurritos This stuff has been addressed

PR #485, opened 2017-07-24
feat(derek): landing page refresh

amberRrucker:
RSWEB-9322

WILL NOT BE PUSHED INTO ZOOLANDER!!!!
Needs to only be reviewed to be put into a general on page and body swapped. All scss and everything will be compiled and put into the body swap region.

Can be viewed here:
http://localhost:3000/derek/templates/landing-page/lockedContent-pdf
and
http://localhost:3000/derek/templates/landing-page/lockedContent-form


amberRrucker:
Overwritten by new LP task

PR #487, opened 2017-07-31
style(derek): adjust events colors

amberRrucker:
RSWEB-9320
Adjusting the colors on the internal events page to remove teal.
Visible here:
http://localhost:3000/derek/view-templates/events/events


runrog:
👍 very good captain Rucker

PR #488, opened 2017-07-31
style(derek): Placing resource teaser links in consistent location

amberRrucker:
RSWEB-9321

Adjusting the resource teaser link to sit at the bottom of the wrapper consistently.
Visible here:
http://localhost:3000/derek/view-templates/resource-center/resource-center

### DIFF
--- a/styleguide/derek/pattern-components/resourceCenter-resource/_resourceCenter-resource.scss
+++ b/styleguide/derek/pattern-components/resourceCenter-resource/_resourceCenter-resource.scss
@@ -58,10 +58,13 @@
 .resource-descWrapper {
   @include make-xs-column(12);
   margin-top: 15px;
+  padding-bottom: 25px;
 }
 
 .resource-linkWrapper {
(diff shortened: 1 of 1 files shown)


styleguide/derek/pattern-components/resourceCenter-resource/_resourceCenter-resource.scss
```diff
@@ -58,10 +58,14 @@
...
   @include make-xs-column(12);
   margin-top: 15px;
+  padding-bottom: 25px;
 }
 
 .resource-linkWrapper {
   @include make-xs-column(12);
+  bottom: 5px;
+  padding-bottom: 10px;
+  position: absolute;
```
WidgetsBurritos: If we're changing this to an absolutely positioned element, is the `@include make-xs-column(12);` stuff still needed?  Seems like to the two are at odds with each other.

amberRrucker:
@WidgetsBurritos sorry. fixed.

PR #490, opened 2017-08-02
fix(derek): Adjust condensed resource

amberRrucker:
@writes here is the quick fix for the condensed resource.


writes:
@amberRrucker thanks!

amberRrucker:
good call @runrog adjusted that.

writes:
+1 LGTM

writes:
@runrog can we get a quick Design Review?

amberRrucker:
+1 to @runrog commit.

PR #493, opened 2017-08-03
feat(tracking): Add call cta tracking

xinbin:
Add different event for call trigger.

### DIFF
--- a/styleguide/derek/global-components/tracking/tracking.js
+++ b/styleguide/derek/global-components/tracking/tracking.js
@@ -83,6 +83,17 @@ Zoolander.Tracking = (function Tracking($) {
       });
     });
 
+    $('.track-pageCall').on('click', () => {
+      dataLayer.push({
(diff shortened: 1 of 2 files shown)


styleguide/derek/global-components/tracking/tracking.js
```diff
@@ -28,6 +28,17 @@ Zoolander.Tracking = (function Tracking($) {
       });
     });
 
+    $('.track-ceilingCall').on('click', () => {
```
WidgetsBurritos: So looking at this a little bit closer, it looks like we're doing double tracking now.  We should remove the `ga.event` that is happening above, and then just modify the `rs.call_click` in GTM to pass in the `transport` value and custom dimensions. See `GA - Generic Event` inside GTM for reference.

xinbin:
@WidgetsBurritos - hey I made the modifications , check the jira ticket for the preview link.

PR #498, opened 2017-08-22
Deprecate resource code

xinbin:
Remove additional-resource and business-resource styles.


styleguide/derek/incubation/solutions.ejs
```diff
@@ -37,7 +37,6 @@
 <%- partial('../pattern-components/ctas/cta.ejs') %>
 <%- partial('../pattern-components/quotes/quote-split.ejs') %>
 <%- partial('../pattern-components/awards/awards.ejs') %>
-<%- partial('../pattern-components/additional-resources/additional-resources.ejs') %>
 <%- partial('../pattern-components/business-resources/business-resources.ejs') %>
```
WidgetsBurritos: Not sure if this is still in WIP or not, but are we gonna be getting rid of the business resources as well?

PR #500, opened 2017-08-28
chore(global): deprecate rug

amberRrucker:
RSWEB-9501 

Removing rug from code base, including tracking.
Can be seen on any incubation or example page.
Should ensure that the tracking remains valid for all other events.


styleguide/derek/global-components/tracking/tracking.spec.js
```diff
@@ -158,6 +133,7 @@ describe('Zoolander Tracking Module', () => {
         eventNonInteraction: 0,
       };
       expect(expected).to.eql(window.dataLayer.pop());
+      result = window.dataLayer.pop();
```
WidgetsBurritos: What's with the new `pop()` here?

amberRrucker:
@WidgetsBurritos I addressed your feedback on this one.

brantdarilek:
👍

PR #501, opened 2017-08-30
feat(slteaser): adding special case for non/linked cards

runrog:
screen shot 2017-08-30 at 1 20 17 pm
See new pattern changes here:

service level teaser (3 up) shows link example,
the others are not linked

http://rsweb-9592-add-nonlink-slteasers.styleguide.dev.website.rackspace.com/derek/incubation/solutions

### DIFF
--- a/styleguide/derek/pattern-components/service-level-teaser/_service-level-teaser.scss
+++ b/styleguide/derek/pattern-components/service-level-teaser/_service-level-teaser.scss
@@ -9,6 +9,10 @@
   position: relative;
   text-decoration: none;
   width: 100%;
+}
+
+// only add a hover class if a link
+.serviceLevelTeaser-link {
 
   &:hover {
     background-color: $gray-lightest;
--- a/styleguide/derek/pattern-components/service-level-teaser/_partials/teaser-link.ejs
+++ b/styleguide/derek/pattern-components/service-level-teaser/_partials/teaser-link.ejs
@@ -15,7 +15,7 @@
   var standoutClass = "";
 } %>
 <!-- Service Level Teaser Pattern -->
-<div class="serviceLevelTeaser <%- standoutClass %>" href="javascript:void(0)">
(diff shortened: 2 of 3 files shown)


styleguide/derek/pattern-components/service-level-teaser/_service-level-teaser.scss
```diff
@@ -9,10 +9,15 @@
...
   text-decoration: none;
   width: 100%;
+}
 
-  &:hover {
-    background-color: $gray-lightest;
-    text-decoration: none;
+// only add a hover class if a link
+a {
+  &.serviceLevelTeaser {
```
WidgetsBurritos: We should be using a separate class for links, instead of attaching to an anchor.  Adding a separate `.serviceLevelTeaser-link` would make sense.

runrog:
changed

amberRrucker:
Could you add a description to the PR? Like 
RSWEB-9592 
Here's where you can find these changes
Here's what I did
Here's any functionality that should be tested

runrog:
@amberRrucker yes sorry, it's updated

styleguide/derek/pattern-components/service-level-teaser/_service-level-teaser.scss
```diff
@@ -9,6 +9,15 @@
...
   width: 100%;
+}
+
+.serviceLevelTeaser {
+  @include service-level-card;
+}
+
+// only add a hover class if a link
+.serviceLevelTeaser-link {
+  @include service-level-card;
```
WidgetsBurritos: Sorry if I didn't clarify before.   I don't think we need the mixin here.  I'd say leave `.serviceLevelTeaser` as-is.  
Then that additional class `.serviceLevelTeaser-link` only has the link-specific behavior, and then in the HTML we would have something like this:

```
<a class="serviceLevelTeaser serviceLevelTeaser-link" href="#"></a>
```

PR #503, opened 2017-08-31
bug(links): Links in bullet lists need to look like links.

xinbin:
Links in a bullet list need to look like links.

### DIFF
--- a/styleguide/derek/pattern-components/leads/_lead.scss
+++ b/styleguide/derek/pattern-components/leads/_lead.scss
@@ -1,9 +1,9 @@
 .lead-bodyWrapper {
   margin-bottom: 14px;
-}
 
-.lead-bodyWrapper > a {
-  @include standard-link;
+  a {
+    @include standard-link;
+  }
 }
 
 .lead-header {
--- a/styleguide/derek/global-theme/photography.jade
+++ b/styleguide/derek/global-theme/photography.jade
@@ -62,8 +62,8 @@ hr
       li This banner style should be used on all pages
       li Multi-line banner text is preferred
       li Banner text should match/relate to the URL, navigation item you click on to get to the page, and/or ay related display ad text
-      li Images must be from the approved list located
-        a(href="https://rax.io/banner-images").red  here
+      li Images must be from the approved list located 
+        a(href="https://rax.io/banner-images").red here
 
 
 .row.half-padding-full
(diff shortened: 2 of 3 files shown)


xinbin:
Current solution affects tabs .. moving on... wip

styleguide/_themes/global/scss/fonts.scss
```diff
@@ -49,6 +49,10 @@ p > a {
   @include standard-link;
 }
 
+li > a {
+  @include standard-link;
+}
```
WidgetsBurritos: image

This seems to set the tab color to red as well.  

I don't think applying to `li > a` is a good idea.  We use `<li>` tags in way too many places.
I think a slightly safer idea would be to target li's inside specific content types where this is a problem. 

Unfortunately that requires undesirable nesting, but that's what the `zoolander/styleguide/_themes/global/scss/bootstrap_overrides.scss` file is for.

xinbin:
[run uat]

xinbin:
Example page: 
hipaa_compliant_hosting___backed_by_baa___rackspace
http://rsweb-9469-style-links.zoolander.dev.website.rackspace.com/certification/hipaa

styleguide/derek/pattern-components/leads/_lead.scss
```diff
@@ -1,5 +1,10 @@
 .lead-bodyWrapper {
   margin-bottom: 14px;
+
+  a {
+    color: $rackspace-red;
+    text-decoration: underline;
+  }
```
WidgetsBurritos: If this is the route we're taking for this, it might make sense to just modify the `.lead-bodyWrapper > a` below to just be `.lead-bodyWrapper a` (i.e. change from the child combinator to the descendent combinator). 

The `standard-link` mixin has all the behavior we want anyway.  Apart from that, I this should be good (pending a thorough site review).

xinbin: I thought better not to jack with existing rules, but aye.

WidgetsBurritos: In this case, it wouldn't actually modify the existing rule, other than making it more permissive, which has the exact same effect as the code you've provided above (just with a few other standard link caveats).

styleguide/derek/pattern-components/leads/_lead.scss
```diff
@@ -1,9 +1,11 @@
 .lead-bodyWrapper {
   margin-bottom: 14px;
-}
 
-.lead-bodyWrapper > a {
-  @include standard-link;
+  a {
+    @include standard-link;
+    color: $rackspace-red;
+    text-decoration: underline;
```
WidgetsBurritos: So I already approved this, but just realized `@standard-link` already provides the `color` and `text-decoration` so those two lines are unnecessary. 
https://github.com/rackerlabs/zoolander/blob/733794300d6bd0322e82f09af08d973c8f2312fe/styleguide/_themes/global/scss/fonts.scss#L37-L46

bighappyface:
[run ci]

PR #504, opened 2017-09-01
feat(homepage): adding new announcment slant section

runrog:
New announcement section of homepage added. Can be seen in zoolander here:

screen shot 2017-09-01 at 1 44 53 pm


styleguide/derek/incubation/homepage.ejs
```diff
@@ -1,4 +1,52 @@
...
+						  <p class="awards-subhead">Hosting engineers across Linux, Windows and VMware</p>
+						  <p class="awards-subtext">including 122 CCNAs and 44 CCNPs</p>
+						</div>
+				  </div>
+				  <div class="col-sm-4">
+						<!-- Awards and Certs Pattern -->
+						<div class="awards">
+						  <h2 class="awards-title">300+</h2>
+						  <p class="awards-subhead">Content Management Specialists</p>
+						  <p class="awards-subtext">including 122 CCNAs and 44 CCNPs</p>
```
amberRrucker: Do you mind tossing in the awards partial here instead? I don't think it's super important to have the content accurate.

runrog: @amberRrucker will do

styleguide/_themes/derek/scss/snowflakes/homepage.scss
```diff
@@ -1,3 +1,98 @@
+.rsHomepage-announcement {
+  display: flex;
+  margin-top: -70px;
+  width: 100%;
+}
+
+.rsHomepage-annLeft {
+  background-image: url('/imgs/home-pageAnn-left.png');
```
WidgetsBurritos: This needs to be a relative path.  (e.g. `../imgs/home-pageAnn-left.png`)
Otherwise this will only ever work on zoolander.

styleguide/_themes/derek/scss/snowflakes/homepage.scss
```diff
@@ -1,3 +1,98 @@
...
+    font-family: 'rswebfonts';
+    font-size: 50px;
+  }
+}
+
+// Desktop only rules
+@media (min-width: $screen-md-min) {
+
+  .rsHomepage-annRight {
+    background-image: url('/imgs/home-pageAnn-right.png');
```
WidgetsBurritos: Also needs to be relative path.

WidgetsBurritos: image
I'm glad this is loading on desktop only, but this image is 799KB and takes a whole second to load. This seems like an unnecessary performance burden for a small component, especially given that this is for our homepage. 

Are there CSS-only mechanisms for making these angles instead of relying on images?

runrog: I have tried css clip-path and skew to get the angel but with too many cons. The clip-path style we need does not work in FF or IE, and skewing the div is a pain to maintain since it breaks different on different screen sizes. I have found that the safest way to do this (cross-browser) and keep the angle intact the best is to use these images.

WidgetsBurritos: We will need to try to crash this then, because adding a second to the load time doesn't seem like a "safe" alternative. It could have negative effects on traffic.

runrog: I think we can condense that size down in photoshop. we can compress it

runrog: Let me see how much I can do

WidgetsBurritos: At the very least, if we do have to use images can we potentially split it into multiple images.  JPG for the majority of the image, and PNG for just the part that needs the angle?

WidgetsBurritos: ^ Not sure how weird that will look

styleguide/_themes/derek/scss/snowflakes/homepage.scss
runrog: I'll change this but this placeholder is likely going to be overwritten in the general with an inline style tag. It was only supposed to work in zoolander.
ie:
```
<div class="rsHomepage-annLeft" style="background-image: url(the_asset_url);">
```

WidgetsBurritos: Where is that asset gonna live?  To me it seems like home page changes warrant a code/qe review anyway due to visibility, so managing that in zoolander seems like the optimal solution here.

runrog:
@WidgetsBurritos updated. Also got the image down to 74.6kb

WidgetsBurritos:
Just to confirm, this is the desired behavior on mobile? 

image

WidgetsBurritos:
For really wide screens, is the content supposed to span all the way to the left like this?  

It's not consistent with the rest of our left margins (except for the tabs, but that's been reported as a "bug"). 

image

WidgetsBurritos:
@runrog Btw, that image is much better. Thank you. I'm honestly surprised that an photo works so well as an SVG.

WidgetsBurritos:
There seems to be a weird artifact on the left hand side.
image

WidgetsBurritos:
I also see some weird artifacts are random screen sizes on both sides of the right column: 
image

runrog:
@WidgetsBurritos 
-the mobile version is the intended look. the top portion that's being cut off won't be there since we are working on the new header, also we are hiding the right side on mobile.

-The left text was also intended to stay intact like that. I'll let @amberRrucker give a second opinion there.

-looks like the tiny slithers of space on each side are because of the "background-size: cover" rule, it's not quite "covering" the entire div. Looking into it.

runrog:
@WidgetsBurritos was finally able to work on this again. So it seemed like the much lower compressed svg/jpg image was causing the edges of the image to become somewhat transparent (even having the image file bleed off the board in photoshop). So I compressed the original PNG down to about 220KB and I'm not seeing any issues on the sides, and the center line is hardly noticeable. Can you give a review please...

WidgetsBurritos:
@runrog, I'm still hesitant signing off on a 220kb image here as it's just one small component on a much larger page.  

Can we try one more thing? Can we split it into two images? 

The top right would be a transparent PNG with an angle like this: 
image

The bottom right would be a higher compressed JPG that is a standard rectangle looking like this: 
image

Then we just make sure they scale together. I'm not sure what that would look like but may offer the ability to compress these images a bit more. But only if it's a meaningful amount of savings in terms of compression as it does add an extra connection.

runrog:
Let me try that out and get back to you

runrog:
@WidgetsBurritos this is ready for review again. image is now a jpeg under 100kb. remember that ff and ie will not have the slant.

WidgetsBurritos:
At various widths, such as 1268px I'm still seeing some bleed through from the banner:
image

WidgetsBurritos:
Also leaving off the clipping in IE/Firefox is fine.  But I'm definitely gonna need to see this in combination with the actual banners to be 100% confident this will work as expected.

That's more of a concern for PR #508 than this one, but I'm just annotating that because there may be some additional tweaks that need to be made around this.

runrog:
@WidgetsBurritos There's just no getting around that gap without dirty hacks. I bled the image off canvas and it still shows. It's just a 1px gap that's not on every size and is barely noticeable. I think we should just leave that as is, otherwise we're going to need to create negative margins and overflows to try and cover it up.

amberRrucker:
@WidgetsBurritos @runrog We can run it through croux with the merge of the homepage banner and see how it goes. Personally, I think it's a non issue.

amberRrucker:
+1 - lets get these together

PR #508, opened 2017-09-06
RSWEB-9534: New Homepage Banner

brantdarilek:
For review:
http://rsweb-9534-homepage.styleguide.dev.website.rackspace.com/derek/incubation/homepage

### DIFF
--- a/styleguide/_layouts/default.jade
+++ b/styleguide/_layouts/default.jade
@@ -96,3 +96,4 @@ html(lang="en")
     $(".imacBanner-headline").fitText(1.1, { maxFontSize: '45px' });
   script.
     $(".refresh-headline").fitText(0.75, { maxFontSize: '55px' });
+    $(".refresh-headline").not('.refresh-headlineHomepage').fitText(0.7, { maxFontSize: '55px' });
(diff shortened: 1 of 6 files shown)


WidgetsBurritos:
In mobile, the top box is slightly larger than the bottom box.  
image

For really wide browsers, the boxes stack on top of each other. 
image

Is this for sure the content here?  Basically, I just want to make sure, if the content was ever 2 lines or 4 lines instead, does the position of the container still look okay? It looks fine to me, but I just wanna make sure ya'll are okay with the alternatives. 
image

WidgetsBurritos:
Is this too much padding on mobile? This is iphone 6 view.  Seems like the CTA shouldn't be below the fold here, but maybe that's just me.  
image

brantdarilek:
@amberRrucker can I get your feedback on the spacing questions?

brantdarilek:
`In mobile, the top box is slightly larger than the bottom box.`
@WidgetsBurritos the boxes are the exact same size what is different is the padding around the text going from 3 lines to 2 lines. @amberRrucker is this the desired look?

WidgetsBurritos:
@brantdarilek,

No if you look closely at that screenshot again, there are two extra pixels on the top box, vs the smaller box. Also I just noticed on that screenshot the angled lines look a little off too: 
image
image

That said, can you confirm the code you just pushed up is correct, because now I'm seeing the boxes next to each other and mobile is just getting cut off. I'm not sure what's up with that, so I couldn't confirm the above was still an issue or not. 
image
image

brantdarilek:
@WidgetsBurritos the code push was not correct. I've since made more adjustments with @amberRrucker on mobile. Would you mind reviewing once more. Thanks

PR #509, opened 2017-09-07
feat(derek): add gated content

amberRrucker:
RSWEB-9573

Added gated content layout.
Also adjusted the callout to be more efficient and less spaghetti.

Can be viewed here:
http://localhost:3000/derek/view-templates/resource-center/pdf-gated

and pages effected by callout adjustment:
http://localhost:3000/derek/templates/landing-page/landingpage-form
http://localhost:3000/derek/templates/landing-page/landingpage-chat

### DIFF
--- a/styleguide/_themes/global/scss/vendor_prefixes.scss
+++ b/styleguide/_themes/global/scss/vendor_prefixes.scss
@@ -88,6 +88,29 @@
   animation: fadein $duration;
 }
 
+@mixin text-fade($color) {
+  background: -webkit-linear-gradient(
+      transparent,
+      $color 90%
+    );
+    background-image: -moz-linear-gradient(
+      transparent,
+      $color 90%
+    );
+    background-image: -o-linear-gradient(
+      transparent,
+      $color 90%
+    );
+    background-image: linear-gradient(
+      transparent,
+      $color 90%
+    );
+    background-image: -ms-linear-gradient(
+      transparent,
+      $color 90%
+    );
+}
+
 @mixin gradient($direction, $firstColor, $secondColor) {
   background: $firstColor;
   background: -webkit-linear-gradient($direction, $firstColor, $secondColor);
--- a/styleguide/derek/view-templates/resource-center/_partials/resource-gated.ejs
+++ b/styleguide/derek/view-templates/resource-center/_partials/resource-gated.ejs
@@ -1,0 +1,29 @@
+<div class="resourcecenter-background half-padding-full">
+  <div class="container">
+    <div class="resourcecenter-fullWrapper resourcecenter-<%- type -%>">
+      <div class="<%- (type == 'video') ? 'col-sm-8 col-sm-push-2' : 'hidden-xs panel-flex col-md-4 col-sm-6' %> text-right">
+        <div class="resourcecenter-visualWrapper">
+        <% if(typeof image != 'undefined') { %>
+          <div class="resourcecenter-imageWrapper" >
+            <img class="resourcecenter-image" src="<%-image%>"/>
+          </div>
+        <% } %>
+        </div>
+      </div>
+      <div class="<%- (type == 'video') ? 'col-sm-8 col-sm-push-2' : 'col-md-8 col-sm-6 col-xs-12' %>">
+        <div class="resourcecenter-infoWrapper">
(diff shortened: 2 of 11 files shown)


WidgetsBurritos:
Some viewports have the callout sitting on top of the icon: 
image

On tablet it gets pushed below the line.  Not sure if this is intentional or not? 
image

WidgetsBurritos:
Another concern is when the form gets longer this looks really funky (this is a concern for regular landing pages as well):
image
image

Forms need to be able to support captchas and what not.  But overall the callouts need to be able to support any length of content.  Short, medium and long.

styleguide/_themes/global/scss/vendor_prefixes.scss
```diff
@@ -88,6 +88,29 @@
...
+      $color 90%
+    );
+    background-image: linear-gradient(
+      transparent,
+      $color 90%
+    );
+    background-image: -ms-linear-gradient(
+      transparent,
+      $color 90%
+    );
```
WidgetsBurritos: can you fix the indentions above?

styleguide/derek/view-templates/resource-center/_partials/resource-gated.ejs
```diff
@@ -0,0 +1,29 @@
...
+        <% if(typeof image != 'undefined') { %>
+          <div class="resourcecenter-imageWrapper" >
+            <img class="resourcecenter-image" src="<%-image%>"/>
+          </div>
+        <% } %>
+        </div>
+      </div>
+      <div class="<%- (type == 'video') ? 'col-sm-8 col-sm-push-2' : 'col-md-8 col-sm-6 col-xs-12' %>">
+        <div class="resourcecenter-infoWrapper">
+          <div class="resourcecenter-titleFull"><h1><%- title %></h1></div>
```
WidgetsBurritos: Instead of nested the H1 in a div can this just be:
```<h1 class="resourcecenter-titleFull"><%- title %></h1>```

**Comment by amberRrucker** [comment]
@WidgetsBurritos , I'm adding in the captcha and legal disclaimer but there is no way there will ever be a download or lead form with that many fields. I'll make sure in the documentation to note a field limit.

File: styleguide/derek/view-templates/resource-center/_partials/resource-gated.ejs (hunk 2 above)
Comment (amberRrucker) [reply styleguide/derek/view-templates/resource-center/_partials/resource-gated.ejs]: This is in all the new resource types. Do you want me to change it for all of them? I don't want them to be inconsistent. That would require a separate www change.

**Comment by amberRrucker** [comment]
@WidgetsBurritos I've addressed your feedback.

File: styleguide/derek/view-templates/resource-center/_partials/resource-gated.ejs (hunk 2 above)
Comment (WidgetsBurritos) [reply styleguide/derek/view-templates/resource-center/_partials/resource-gated.ejs]: ehh just leave it as-is for now then.

File: styleguide/derek/pattern-components/lp-callout/_partials/lp-callout-form.ejs [hunk 3]
```diff
@@ -17,5 +17,7 @@
       <input type="phone" class="form-control" id="phone">
     </div>
   <button type="submit" class="button lead calloutForm-button">Send</button>
+  <img class="calloutForm-captcha" src="/imgs/derek/captcha.jpg">
```
WidgetsBurritos: Our current version of recaptcha doesn't scale and mandates dimensions of 304px width and 78px height. 

We can simulate this behavior in zoolander by adding this to the image: 
`style="width: 304px; height: 78px; max-width: none;"`

What we see is that in tablet views, the captcha does not fit: 
image

So we have one of two options here: 
1. Resize this box
2. Modify recaptcha similar to what they do here: https://geekgoddess.com/how-to-resize-the-google-nocaptcha-recaptcha/

That said, option 2 has implications reaching far past gated content and landing pages.  It would impact forms across the board and it's hard to say if it would work properly or not. I'd prefer option 1 as it limits the scope, but that's probably a decision croux will need to make.

runrog:
What happened to the CTA box edging into the top header area? The non form one is not doing it anymore in any viewport. 

screen shot 2017-09-20 at 12 30 44 pm

amberRrucker:
@runrog got lost in a merge. its back!

PR #514, opened 2017-09-14
RSWEB-9665: Fix google plus social button

brantdarilek:
This PR fixes an issue with google plus button adding more width then necessary with inline styles dynamically in safari. 

Unfortunately, there is not class added to the dynamically created button. So I had to disable a couple of rules, only for the social.scss file.

### DIFF
--- a/styleguide/derek/global-components/social/_social.scss
+++ b/styleguide/derek/global-components/social/_social.scss
@@ -3,6 +3,15 @@
   top: -8px;
 }
 
+.social-button {
+  display: inline-block;
+  margin: 0 2px;
+}
+
+.social-googleplusWrapper {
+  width: 57px;
+}
+
 .social-shareEmail {
   background-color: $gray-base;
   border-radius: 3px;
--- a/styleguide/derek/global-components/social/social.ejs
+++ b/styleguide/derek/global-components/social/social.ejs
@@ -18,8 +18,18 @@
 <script src="https://apis.google.com/js/platform.js" async defer></script>
 
 <!-- Button Codes -->
-<script type="IN/Share" data-url="rackspace.com"></script>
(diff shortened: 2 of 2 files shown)


xinbin:
I trust that using important was the only viable solution?

brantdarilek:
Unfortunately, yes. This is what the script adds
`<div style="text-indent: 0px; margin: 0px; padding: 0px; background: transparent; border-style: none; float: none; line-height: normal; font-size: 1px; vertical-align: baseline; display: inline-block; width: 93px; height: 20px;" id="___plus_0">`

styleguide/derek/global-components/social/_social.scss
```diff
@@ -1,8 +1,15 @@
...
+
 //Hack for Facebook to work.
 .fb-share-button {
   top: -8px;
 }
 
+//Hack for Google Plus.
+#___plus_0 {
+  width: 57px !important;
+}
```
WidgetsBurritos: Since we control all the markup around these elements can we try something like this instead?

```html
<div class="social-button">
    <script type="IN/Share" data-url="rackspace.com"></script>
</div>
<div class="social-button">
    <div class="fb-share-button" data-href="<?php print $path; ?>" data-layout="button" data-size="small" data-mobile-iframe="true"><a class="fb-xfbml-parse-ignore" target="_blank" href="https://www.facebook.com/sharer/sharer.php?u=<?php print rawurlencode($path); ?>&amp;src=sdkpreparse"><?php print t('Share', [], ['context' => 'Social Buttons']); ?></a></div>
</div>
<div class="social-button">
    <a href="https://twitter.com/intent/tweet?text=<?php print $twitter_text; ?>" class="twitter-share-button" data-show-count="false"><?php print t('Tweet', [], ['context' => 'Social Buttons']); ?></a>
</div>
<div class="social-button">
    <div class="g-plus" data-action="share"></div>
</div>
<div class="social-button">
    <a class="social-shareEmail" href="mailto:?subject=<?php print $email_subject; ?>&body=<?php print $email_body; ?>" ><?php print t('Email', [], ['context' => 'Social Buttons']);?></a>
</div>
```

Then we can mandate with from an element above the actual social button. 

I'd just rather not use IDs or `!important` if possible. I did an initial test of this in the safari element inspector and it seems to work (although I admittedly haven't throughly tested it).

xinbin:
+1 ty

runrog:
+1 Looks good

screen shot 2017-09-18 at 10 21 59 am

PR #516, opened 2017-09-19
refactor(derek):banner rework

amberRrucker:
RSWEB-9704

Reviewing Brant's changes. Go Brant!

Visible in these locations:
http://rsweb-9541-banner-rework.styleguide.dev.website.rackspace.com/derek/templates/landing-page/landingpage-chat

http://rsweb-9541-banner-rework.styleguide.dev.website.rackspace.com/derek/view-templates/leadership/leadership

http://rsweb-9541-banner-rework.styleguide.dev.website.rackspace.com/derek/incubation/homepage

Keep in mind there is a little blip when it first loads the page because of the animation of the zoolander side navigation.

### DIFF
--- a/styleguide/derek/pattern-components/headers/_headers.scss
+++ b/styleguide/derek/pattern-components/headers/_headers.scss
@@ -1,3 +1,5 @@
+$red-variation : #c32126;
+
 //used for events
 @mixin make-banner {
   background-repeat: no-repeat;
@@ -159,67 +161,51 @@
   }
 }
 
-
-
 .refreshBanner {
   @include make-banner;
+  @include fadein(2s);
   background-image: url('../imgs/BG.jpg');
-  background-position: center right;
-  height: 515px;
+  background-position: right center;
+  height: 300px;
   overflow: hidden;
+  padding: 60px 0;
+  position: relative;
+}
+
+.refreshBanner-landingPage {
+  height: 60vw;
+  max-height: 515px;
   padding: 150px 0 0;
 }
 
+.refreshBanner-homepage {
+  height: 60vw;
+  max-height: 620px;
+  padding-top: 90px;
+}
+
 .refreshBanner-container {
-  height: 100%;
+  @include make-container;
+  @include fadein(1s);
+  padding: 0 45px;
   position: relative;
-  width: 100%;
 }
 
 .refreshBanner-boxContainer {
   height: 100%;
   padding: 0;
   position: relative;
-}
-
-.refreshBanner-boxContainer2 {
-  clear: both;
-  top: -65.5%;
-}
-
-.refreshBanner-boxTop {
-  display: table;
-  min-height: 40%;
   width: 100%;
 }
 
-.refreshBanner-boxTopInterior {
-  min-height: 40%;
+.refreshBanner-boxContainerBottom {
+  clear: both;
+  position: absolute;
+  top: 32%;
+  width: 70%;
 }
 
-.refreshBanner-boxTopHomepage {
-  min-height: 10%;
-}
-
-.refreshBanner-innerTopLeft {
-  @include make-angle;
-  background-image: url('../imgs/slant-topleft.svg');
-  width: 9.6667%;
-}
-
-.refreshBanner-innerTopRight {
-  @include make-angle;
-  background-color: $white;
-  padding: 25px;
-  vertical-align: middle;
-  width: 83.3333%;
-}
-
-.refreshBanner-innerTopRight2 {
-  width: 79.3333%;
-}
-
-.refresh-headline {
+.refreshBanner-headline {
   color: $brand-primary;
   font-family: $heading;
   font-size: 6rem;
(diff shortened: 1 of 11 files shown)


styleguide/derek/pattern-components/headers/_headers.scss
```diff
@@ -229,139 +215,223 @@
...
+  background-color: $white;
+  display: block;
+  z-index: 0;
 
-.refreshBanner-sizeMedium {
-  height: 400px;
-  padding-top: 100px;
-}
+  &.left-top {
+    // -webkit-transform: skewY(-45deg);
```
WidgetsBurritos: can this can be removed?

styleguide/derek/pattern-components/headers/_headers.scss
```diff
@@ -229,139 +215,223 @@
...
+    width: 300px;
+  }
+
+  &.left-topHomepage {
+    top: 300px;
+  }
+
+  &.right-bottom {
+    @include make-sm-column(7);
+    // -webkit-transform: skewX(-45deg);
```
WidgetsBurritos: can this be removed?

styleguide/derek/pattern-components/headers/headers.js
```diff
@@ -0,0 +1,44 @@
+Zoolander.Banners = (() => {
+  function interiorBanner() {
+    const $textContainer = $('.refreshBanner-textContainer');
+    const $leftSlant = $('.refreshBanner-line.left-top');
+    const $mainContainer = $('.refreshBanner-container');
+    const $mainContainerLeft = $mainContainer.offset().left;
```
WidgetsBurritos: the `$` in front of a variable implies you're working with a jQuery object. While true in the three of the cases above this, in this particular case it is not true.  I'd remove the `$` on any variables that aren't jQuery objects to provide better clarity. This applies to all variables in this file.

styleguide/derek/pattern-components/headers/headers.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    });
+
+    it('should adjust the left angle', () => {
+      $('body').append('<div class="refreshBanner"><div class="refreshBanner-container"><div class="refreshBanner-boxContainer"><div class="refreshBanner-textRow"><div class="refreshBanner-textContainer refreshBanner-textContainerInterior"><h2 class="refreshBanner-headline">we guarantee a 99.99% openstack API uptime.</h2></div></div><span class="refreshBanner-line right-bottom">&nbsp;</span></div><span class="refreshBanner-line left-top">&nbsp;</span></div></div>');
+      const offset = $('.refreshBanner-line.left-top').offset();
+      Zoolander.Banners.init();
+      setTimeout(() => {
+        expect(offset).to.not.eql($('.refreshBanner-line.left-top').offset());
+      }, 1000);
+    });
```
WidgetsBurritos: I don't think we need all this fancy behavior in our unit tests. This tests should be written the same way we did the filter bar tests. 

See: https://github.com/rackerlabs/zoolander/blob/master/dist/js/global-components/filter-bar/filter-bar.spec.js

This may mean some slight refactoring of the code.   

What needs to be tested is the actual mathematical calculations.  If we can vouch for inputs and outputs in all situations, that's enough test coverage. The tests as currently written only test whether or not offsets are not equal to one another.  It doesn't matter if they are correct calculations or not, just that they aren't the same value. 

By doing the other method we can get rid of DOM manipulation, the need to initialize code mid test, and get rid of fittext related timeouts (which may or may not flake out depending on how flaky connections decide to be at any given moment).

styleguide/js/global.js
```diff
@@ -0,0 +1,102 @@
...
+  searchSubmit.on('click', function () {
+    if (searchContainer.hasClass('navbar-searchExpanded')) {
+      searchContainer.removeClass('navbar-searchExpanded');
+      searchBox.trigger('blur');
+    } else {
+      searchContainer.addClass('navbar-searchExpanded');
+      searchBox.trigger('focus');
+    }
+  });
+})(jQuery);
```
WidgetsBurritos: Why was this file included in this PR?  Is this necessary?

WidgetsBurritos:
@brantdarilek 

image

Banners on the landing pages are still too small. 
Compare the left side (zoolander) with the right side (production). 

An example of this is: https://www.rackspace.com/dedicated-servers/lp/mssql-bundle

WidgetsBurritos:
@brantdarilek, @amberRrucker,

It's looking better. Are we concerned that this renders slightly differently in safari than in other browsers? 

(Safari is on the far right:)

image

And we probably need to check this in IE.

WidgetsBurritos:
@amberRrucker @brantdarilek Please see the latest commit I pushed to this.  I believe this fixes our problem. What I did was as follows:

1. Modified fittext to fire an event after completion
2. Loaded in the fontfaceobserver library to let us observe when khand/700 is ready.
3. I then use the fontfaceobserver to initiate fittext. (Note that we have two different conditions here.  One is if Khand successfully loads. The other is a fallback in case khand ever doesn't load for some reason. This is important just in case khand were to ever break for whatever reason. Notice I picked a smaller max font as the fallback is a lot larger.) 
4. I split the homepage banner up into two separate functions (one for top, one for bottom).
5. I moved all the jquery selector lookups outside of the banner callbacks, as otherwise the browser would have to look these up every time the browser resizes. Since it's never changing I figured the single lookup was optimal here. 

See Jira ticket for preview link.

brantdarilek:
+1 807ce7c

WidgetsBurritos:
Closing as we're gonna be trying a simpler solution separately.

PR #517, opened 2017-09-21
feat(hexcalc): add in hex calculator code off cdn

xinbin:
Step 1 of many get the hex calculator code off cdn and into zoolander.
This is old code that will not run as it is with 'use strict'. I disabled this file from having that applied to it and had to create a work around to get it added to the code that is distributed.


xinbin:
[run ci]

xinbin:
[run ci]

PR #519, opened 2017-09-25
feat(derek): add nascar river

amberRrucker:
RSWEB-9541

Creating a new nascar river to mirror design linked in the ticket. Only weirdness is that I had to create an override for super small screens to go full width. I also took out legacy example code on the homepage that is no longer there so we can have a proper stakeholder review.

This can be seen here:
http://localhost:3000/derek/incubation/homepage
beneath the announcement section (which isn't the final version but I needed to see it on page)

### DIFF
--- a/styleguide/derek/incubation/homepage.ejs
+++ b/styleguide/derek/incubation/homepage.ejs
@@ -1,145 +1,63 @@
 <%- partial("../pattern-components/headers/header-refresh-homepage.ejs") %>
 
-<div class="container half-padding hostingOptions">
-	<div class="row half-padding-full">
-		<div class="col-md-11 center-block">
-			<h1 class="text-center no-margin no-padding" id="homepage-lead-top">Rackspace is the largest managed cloud provider</h1>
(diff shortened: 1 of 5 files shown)


styleguide/derek/incubation/homepage.ejs
```diff
@@ -1,146 +1,64 @@
...
+	<div class="rsHomepage-leftStripContainer">
+		<div class="rsHomepage-annLeftBar">
+			<div class="rsHomepage-titleContainer">
+				<h2 class="rsHomepage-annLeftBar-header">TRANSFORM THE WAY YOU DO BUSINESS</h2>
+				<h3 class="rsHomepage-annLeftBar-subHeader">Offering expertise and support for your IT transformation</h3>
+			</div>
 		</div>
+		<div class="rsHomepage-annRight-iconBar"></div>
 	</div>
-	<div class="row row-eq-height">
```
amberRrucker: @WidgetsBurritos we discussed this. I had to remove these sections in order to see the homepage as it will be. No styling is removed in this PR. Just some of this random leftover junk.

runrog:
+1 Design

PR #522, opened 2017-10-02
feat(derek): resource refresh documents card

amberRrucker:
RSWEB-9677
Adding the resource center card for documents.

Visible here:
http://localhost:3000/derek/view-templates/resource-center/refresh/resource-center

Couple questions:
I needed to add the image in css. I'm getting an example image from Mike to see if these will be okay to be used as the background in smaller screen sizes. 

There are 3 rgba values in here for shadows. I don't feel like they necessarily need to be swatch values. How do you want these handled? Individual variables in this file?


styleguide/derek/pattern-components/resourceCenter-resource/refresh/_resourceCenter-resource.scss
```diff
@@ -0,0 +1,141 @@
+// Resource Icon Color
+@mixin resource-refresh($class, $color) {
```
WidgetsBurritos: `$color` here implies you're always passing in a color.  Rename this to `$background` as that is more indicative of what you're actually doing.

styleguide/derek/pattern-components/resourceCenter-resource/refresh/_resourceCenter-resource.scss
```diff
@@ -0,0 +1,141 @@
...
+    box-shadow: 0 1px 1px 1px rgba(0, 0, 0, .06);
+    padding: 0;
+    transition: .3s ease-in-out;
+    width: 100%;
+
+    .resource-top {
+      background: $color;
+    }
+
+    a {
```
WidgetsBurritos: We should use a class for here if possible.

styleguide/derek/pattern-components/resourceCenter-resource/refresh/_resourceCenter-resource.scss
```diff
@@ -0,0 +1,141 @@
...
+  line-height: 1.7em;
+}
+
+.resource-link {
+  @include standard-link;
+  font-family: $heading;
+  font-weight: 600;
+  text-decoration: none;
+
+  .fa {
```
WidgetsBurritos: We should use a top-level, maintainable class for this.

styleguide/derek/pattern-components/resourceCenter-resource/refresh/_resourceCenter-resource.scss
```diff
@@ -0,0 +1,141 @@
...
+
+    a {
+      @include flex-box;
+      flex-flow: column nowrap;
+      height: 100%;
+      text-decoration: none;
+    }
+
+    &:hover {
+      box-shadow: 0 15px 25px -9px rgba(0, 0, 0, .2);
```
WidgetsBurritos: Per your question in jira, all colors (even colors with various opacities) need to be in the swatches file.  That's the purpose of the swatches file, to keep us from creating too many color variations throughout the site.

WidgetsBurritos:
Is it intentional that we're showing the title twice on each card? 

image

styleguide/derek/pattern-components/resourceCenter-resource/refresh/_resourceCenter-resource.scss
amberRrucker: We tried this with both iterations of the resource cards so far and removing it breaks all the styling. I just tried it again to confirm and unfortunately that is still the case.

amberRrucker:
@WidgetsBurritos I pushed up the changes you requested but my question was mainly around will this structure work with www development. I want to get started on the other cards but I need to make sure this structure is okay for www.

WidgetsBurritos:
In regards to structure it's fine.  Just some minor cleanup should be necessary.

WidgetsBurritos:
But to clarify my previous statement, I'd rather get this single pattern rock solid before we start dealing with PRs for the other patterns. Since this is all new code it can be merged in when ready making that process easier.

styleguide/_themes/global/scss/swatch.scss
```diff
@@ -36,3 +36,9 @@ $brand-danger:      $rackspace-red !default;
 
 //alpha colors
 $black-shadow:		rgba($black, .175);
+
+$clickablebox-shadow: rgba($black, .06);
+$clickablebox-hover: rgba($black, .2);
+$top-overlay-shadow: rgba($black, .15);
```
WidgetsBurritos: It would be preferable to use a scaled naming convention here like we do for grays above.  
Something like:
`$black-shadow-dark`
`$black-shadow-darker`
`$black-shadow-darkest`
etc...

PR #523, opened 2017-10-04
RSWEB-9655: feat(resources): new side bar filtering style

runrog:
Zoolander link: http://localhost:3000/derek/view-templates/resource-center/resource-center

screen shot 2017-10-04 at 10 21 16 am


WidgetsBurritos:
image

When you first load it is all collapsed, but "Resource Types" is highlighted as if it is expanded.  That's confusing.

styleguide/derek/global-components/filter-bar/filter-accordion.js
```diff
@@ -0,0 +1,64 @@
...
+        $formContent.slideToggle(frameRate);
+      },
+      resizeTimer: null,
+      resetForm() {
+        clearTimeout(this.resizeTimer);
+        this.resizeTimer = setTimeout(() => {
+          if ($window.outerWidth() > 768 && !$formContent.is(':visible')) {
+            $formContent.show();
+          }
+        }, 250);
```
WidgetsBurritos: What is the purpose of this timeout here?

WidgetsBurritos:
This looks really funky at around 780px: 

image

styleguide/derek/global-components/filter-bar/filter-accordion.js
runrog: when you're resizing from mobile to desktop. if you collapsed it in mobile, then move back to desktop, it needs to show up again. The timeout is so that it won't run a million times for every pixel resized

WidgetsBurritos: can we set the timeout to `100` instead of `250` then? The human eye can perceive up to about 1/10 of a second, so it may help reduce jitter, and still prevent the busy waiting there.

styleguide/derek/global-components/filter-bar/filter-accordion.js
```diff
@@ -0,0 +1,64 @@
...
+        $siblings.find(contentClass).slideUp(frameRate);
+      },
+      toggleForm() {
+        $formContent.slideToggle(frameRate);
+      },
+      resizeTimer: null,
+      resetForm() {
+        clearTimeout(this.resizeTimer);
+        this.resizeTimer = setTimeout(() => {
+          if ($window.outerWidth() > 768 && !$formContent.is(':visible')) {
```
WidgetsBurritos: Should this be `>= 768`?  We change viewports at exactly 768px so it seems we want to be inclusive of that.

runrog:
@WidgetsBurritos feedback addressed

WidgetsBurritos:
@runrog The text wrapping still looks funky at around 780px.   

image

Additionally the button text is expanding outside of the button: 

image


I'm wondering if we should just use the mobile view for the 768px - 991px breakpoint as well?

styleguide/derek/global-components/filter-bar/_filter-bar.scss
```diff
@@ -138,3 +138,163 @@
     margin: .5rem auto 0;
   }
 }
+
+// Start of accordion style side bar
+.rsFilter-sideBar {
+  @include make-sm-column(2);
```
WidgetsBurritos: Is 2 columns too narrow here? We're all but guaranteed to get wrapping text like this: 

image

I realize that may require a bit of coordination with @amberRrucker on the main resource content area as well, but I just feel like we're trying to cram too much text in too small of a region.

runrog: This is the current behavior of the side bar. It's a col (2)

WidgetsBurritos:
Alternatively can we get it to wrap like this?

image

styleguide/derek/global-components/filter-bar/_filter-bar.scss
WidgetsBurritos: If this task is to cleanup the sidebar though, I don't necessarily see solving this as outside of the scope.

runrog:
@WidgetsBurritos I went with the no wrap approach for the labels. Ran it by CROUX, and 👍 on it

runrog:
feedback addressed. now breaking at 991px instead of 760, labels are now broken in 2 lines

styleguide/derek/global-components/filter-bar/filter-accordion.js
```diff
@@ -0,0 +1,69 @@
...
+        $siblings.find(contentClass).slideUp(frameRate);
+      },
+      toggleForm() {
+        $formContent.slideToggle(frameRate);
+      },
+      resizeTimer: null,
+      resetForm() {
+        clearTimeout(this.resizeTimer);
+        this.resizeTimer = setTimeout(() => {
+          if ($window.outerWidth() >= 991 && !$formContent.is(':visible')) {
```
WidgetsBurritos: Should this be either `> 991` or `>= 992` instead?  Our breakpoint happens at 992, not 991.

runrog:
@WidgetsBurritos done

WidgetsBurritos:
[run ci]

PR #524, opened 2017-10-04
feat(derek): add code resource type

amberRrucker:
RSWEB-9718
Adding new code resource type to the refresh resource types.

Visible with the new document resource here:
http://localhost:3000/derek/view-templates/resource-center/refresh/resource-center


WidgetsBurritos:
[run ci]

PR #536, opened 2017-10-17
Revert "RSWEB-9655: feat(resources): new side bar filtering style"

WidgetsBurritos:
Reverts rackerlabs/zoolander#523

^ accidentally merged this too early.


runrog:
+1

PR #538, opened 2017-10-17
RSWEB-9750: remove tab horizontal scroll

writes:
http://rsweb-9750-tabs-horizontal-scroll.zoolander.dev.website.rackspace.com


WidgetsBurritos:
@amberRrucker, @xinbin,

FYI, the tabs not working appears to be an issue with the PR builder urls here as it can't seem to find tab collapse. The zoolander PRs have a few problems with them that will probably need to be resolved at some point.

PR #539, opened 2017-10-17
RSWEB-9339: resource image size in tablet viewports

writes:
http://rsweb-9339-resource-image.zoolander.dev.website.rackspace.com/resources/whom-can-you-depend

### DIFF
--- a/styleguide/derek/view-templates/resource-center/_resource-center.scss
+++ b/styleguide/derek/view-templates/resource-center/_resource-center.scss
@@ -65,7 +65,7 @@
 
 .resourcecenter-imageWrapper {
   height: 100%;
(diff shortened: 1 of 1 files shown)


writes:
[run ci]

styleguide/derek/view-templates/resource-center/_resource-center.scss
```diff
@@ -228,20 +239,15 @@
   }
 }
 
-@media screen and (min-width: $screen-md-min) {
```
writes: @WidgetsBurritos moved this because we had a duplicate media call

brantdarilek:
+1

PR #540, opened 2017-10-18
pattern(interactiveMap): new interactive pattern for rs dc map

runrog:
Dev Link:
http://rsweb-9711-interactivemap.styleguide.dev.website.rackspace.com/derek/view-templates/interactive-map/map

Had to add in a gauge icon and it looks like the new up/down icons are trying to come in from the config of the resource filtering PR. So this may need to wait for that to be merged (depends on what we want to merge first). 

The map.ejs file is how this would be used in a general.

The tips now move to the side if bleeding too far up the page
screen shot 2017-10-20 at 10 46 51 am


styleguide/derek/view-templates/interactive-map/interactive-map.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  describe('calculate left value off screen', () => {
+    it('should add 15 to the circle width', () => {
+      const data = {
+        offScreen: true,
+        c: 28,
+      };
+      expect($.fn.rsInteractiveMap.test.setLeftCoords(data))
+            .to.eql(data.c + 15);
+    });
+  });
```
WidgetsBurritos: Since this is `setLeftCoords()`, which is also being tested in the test prior to this, I'd place this `it()` statement inside the previous `describe()` block and then modify that describe block to something like `Test setLeftCoords()`. 

`describe()` is a good way to lump similar tests together.  I'd do the same for `setTopCoords()` tests below.

runrog:
@WidgetsBurritos feedback addressed

xinbin:
[run ci]

PR #541, opened 2017-10-20
RSWEB-9536 Interior Banner HTML validation

bighappyface:
This PR adjusts some `headers.js` logic to pass W3C HTML validation when the code is embedded within an HTML document.


writes:
+1

PR #542, opened 2017-10-30
chore(derek):replace announcement image

amberRrucker:
RSWEB-9996
Visible on the homepage in zoolander
http://localhost:3000/derek/incubation/homepage

At this point I basically just need a dev branch so I can see how everyone feels about it. I can't tell if I'm being overly critical of my work or if this image sucks.


runrog:
Closing as this is no longer needed.

PR #546, opened 2017-10-31
RSWEB-9999: Fix padding to better show chat slider

brantdarilek:
Before change:
screen shot 2017-10-31 at 12 07 59 pm

After change:
screen shot 2017-10-31 at 12 07 46 pm

### DIFF
--- a/styleguide/derek/global-components/filter-bar/_filter-bar.scss
+++ b/styleguide/derek/global-components/filter-bar/_filter-bar.scss
@@ -353,3 +353,9 @@
     display: none;
   }
 }
+
+@media (min-width: $screen-sm-min) {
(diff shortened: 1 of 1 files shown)


WidgetsBurritos:
It's still doing it on tablet:
image

I might recommend just setting `padding-right: 50px` on tablet and larger instead.  That way only the right side is affected. Something like this: 

```sass
  .rsFilter-results-content {
    padding-right: 50px;
  }
}
```

brantdarilek:
@WidgetsBurritos 
We don't show the chat slider on tablet devices.

writes:
+1

writes:
@amberRrucker @runrog could you provide a design review?

PR #548, opened 2017-11-01
style(derek): RSWEB-9853 - changed max-width on quote wrappers

runrog:
widening quotes on rivers
screen shot 2017-11-01 at 2 39 41 pm

local link
http://localhost:3000/derek/patterns/patterns


writes:
+1

PR #551, opened 2017-11-15
fix(map): fixing bubble close too fast issues

runrog:
This fixes the close too early issue and it's less sensitive to opening unintended bubbles.


runrog:
@WidgetsBurritos Mike and I have spoken about this but we are gonna need a small CROUX discussion. There isn't actually a delay in it but there is an invisible parent around it for more of a threshold when hovering around it. Like so:

screen shot 2017-11-16 at 2 26 32 pm

runrog:
@WidgetsBurritos can I get another review. we are now using the hoverintent plugin
http://rsweb-10066-map-bubblefollow.styleguide.dev.website.rackspace.com/derek/view-templates/interactive-map/map

PR #552, opened 2017-12-04
RSWEB-9948: Use jQuery UI for modals

WidgetsBurritos:
This PR swaps modals to jQuery UI so we can use the modal system provided by Drupal 8 instead of our own custom system.


styleguide/global/modal.js
```diff
@@ -0,0 +1,95 @@
...
+      'ui-dialog-titlebar-close': 'rackspaceModal-titleBarClose',
+    };
+
+    // Sets proper window position.
+    const positionWindow = () => {
+      const windowWidth = $(window).width();
+      const windowHeight = $(window).height();
+      const modalWidth = windowWidth * 0.95;
+      const modalMaxHeight = windowHeight * 0.8;
+      $('.rackspaceModalSource')
```
runrog: This is mostly good but I'd recommend caching this lookup since you're using it twice. 
```const modal = $('.rackspaceModalSource');```

WidgetsBurritos: @runrog changed

runrog:
My take as far as design: 

1) This is pending discussion in CROUX but I'm feeling the scroll within the box probably isn't the way to go because that scroll bar looks different in different browsers and can be an eyesore. The modal also looks a little funny since content is cut off. It's less of a distraction when the entire overlay has the scroll.
screen shot 2017-12-04 at 10 46 41 am

2) I'm cool with the modal being dead center just as long as they are all consistently like that. Some minor cosmetic changes I'd recommend is to remove the default border-radius to not rounded, widen the modal to around twice the size it is now and finally, to replace that "X"  close button. It's a little too small and light.
screen shot 2017-12-04 at 10 51 41 am

brantdarilek:
LGTM +1

PR #559, opened 2018-02-12
MTE-111: Add redirect modal flag options

WidgetsBurritos:
This PR just adds a copy of the basement flag icons for the redirect modal just so we can get a slightly more maintainable approach in D8.


xinbin:
+1

PR #566, opened 2018-05-16
fix(banners): Resolve non-wrap for banner

xinbin:
https://jira.rax.io/browse/MTE-1958

### DIFF
--- a/styleguide/derek/pattern-components/headers/headers.js
+++ b/styleguide/derek/pattern-components/headers/headers.js
@@ -46,10 +46,10 @@ function svgBannerStringSplitter(str, l){
 /**
  * Set's the svg content.
  */
-function svgBannerSetContent() {
+function svgBannerSetContent(primaryBox = 25) {
   // Add content lines.
   var boxTextLimits = {
-    primaryBox: 25,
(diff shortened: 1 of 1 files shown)


styleguide/derek/pattern-components/headers/headers.js
```diff
@@ -46,10 +46,10 @@ function svgBannerStringSplitter(str, l){
 /**
  * Set's the svg content.
  */
-function svgBannerSetContent() {
+function svgBannerSetContent(primaryBox = 24) {
```
WidgetsBurritos: I don't think we want to change the default value here.  It could have unexpected results.

WidgetsBurritos: To clarify, the default was 25 before, and this changes it to 24.  We should leave the default at 25.

xinbin: There is an issue on D7 and I'm sure D8 where the text butts up to the edge of the white box. Changing it to 24 resolved that issue.. Looks like this:
screen shot 2018-05-14 at 11 47 18 am

PR #574, opened 2018-08-21
feat(derek): add homepage case study section

amberRrucker:
MTE-2747

Adding the case study section to the homepage. Background image and content should be able to be altered within the CMS similar to the banners. 

Viewable here:


styleguide/derek/incubation/homepage.ejs
```diff
@@ -52,20 +42,21 @@
...
-            <div class="quadrant-buttonContainer">
-              <a class="button learning" href="[rsweb:why-rackspace]">Learn More</a>
-            </div>
+<!-- Case Study -->
+
+<div class="caseStudy-wrapper">
+  <div class="caseStudy-container">
+    <div class="caseStudy-row">
+      <div class="caseStudy-content">
+        <img class="caseStudy-image" src="/imgs/derek/logos/mckesson_logo.png"/>
```
WidgetsBurritos: Is it just me or is the McKesson logo blurry?

image

styleguide/_themes/derek/scss/snowflakes/homepage.scss
```diff
@@ -290,6 +290,59 @@
   text-align: right;
 }
 
+//Case Study
+  
+.caseStudy-wrapper {
+  background-image: url('../imgs/derek/banners/SSImage.jpg');
```
WidgetsBurritos: image

Do we care about how the kid's face gets chopped in wider browsers?

styleguide/derek/incubation/homepage.ejs
amberRrucker: Yeah a little bit. I don't have a higher quality logo. Thats the one I got in the task, it's more of a question for Mike like hey do we want a smaller one or do we want to source a better logo. It would be switched out in the CMS anyways so it's not the hard and fast logo.

styleguide/_themes/derek/scss/snowflakes/homepage.scss
amberRrucker: @WidgetsBurritos That's a Mike question again. The image will change depending on the client we feature. To make it work for whatever client I'll need to edit the images individually. So if I adjust where this image sits, the next client image could look terrible. It's just like the banners. It's part of why full image backgrounds arent my favorite thing.

PR #580, opened 2018-10-08
MTE-3099: Swap out font with Fira Sans for analysis

WidgetsBurritos:
This PR tests replacing Open Sans and Khand with Fira Sans.


WidgetsBurritos:
All done.

PR #584, opened 2018-11-13
MTE-3377: Swap to Fira Sans

WidgetsBurritos:
This PR loads pulls in Fira Sans via `npm` and then adds a `postinstall` command to install the font directly into the styleguide directory.  

Additionally I had to convert `harp` to an optional dependency and install the `pify` dependency and update `node-sass`.


WidgetsBurritos:
[run ci]

PR #593, opened 2018-11-20
feat(custstory): Updated customer story page..

xinbin:
zoolander_desktop
desktop_
zoolander_tablet
tablet_
zoolander_mobile
mobile_


WidgetsBurritos:
@xinbin,

The background image seems off from 992px -> 1175px.
image

runrog:
@xinbin you can probably safely target the broken viewport with something like this
```
@media screen and (min-width: $screen-md-min) and (max-width: $screen-lg-min) {
  // img size chage
}
```

styleguide/_themes/derek/scss/snowflakes/homepage.scss
```diff
@@ -331,17 +364,156 @@
...
+
+  .caseStudy-bar {
+    border-top: 1px solid $gray-light;
+    padding-top: 10px;
+  }
+
+  .container {
+    div {
+      :last-child {
+        .caseStudy-box {
```
WidgetsBurritos: We shouldn't ever need to use element-based selectors, and there's a lot of nesting going on here. In these particular cases I'd just give each box a unique name. 
For example, instead of something like this: 
```
<div class="caseStudy-box col-md-3 col-md-offset-6 col-sm-6">
```

let's go with 
```
<div class="caseStudy-box caseStudy-box-story1">
```

Then we just define the css like: 

```
.casestudy-box-story1 {
  @include make-sm-column(6);
  @include make-md-column(3);
  @include make-md-column-offset(6);
}
```

This allows us to simplify the markup and get rid of some of the duplicative bootstrap selectors. 

We'd then need to do that for all of the other boxes as well.

styleguide/derek/incubation/homepage.ejs
```diff
@@ -5,6 +5,38 @@
...
+        </div>
+        <div class="row-eq-height">
+          <div class="caseStudy-box col-md-3 col-md-offset-3 col-sm-6">
+            <div class="caseStudy-bar">
+              <img class="caseStudy-image" src="/imgs/derek/logos/mckesson-blk.png"/>
+              <p class="caseStudy-text">McKesson simplified its dated, complex workflow by migrating data to the cloud, and transformed its management abilities.</p>
+              <a class="caseStudy-button">Read their story &gt;</a>
+            </div>
+          </div>
+          <div class="caseStudy-box col-md-3 col-md-offset-3 col-sm-6">
```
WidgetsBurritos: Per feedback above, let's avoid using standard bootstrap classes here in favor of more maintainable class names.

styleguide/derek/incubation/homepage.ejs
```diff
@@ -5,6 +5,38 @@
...
+            <div class="caseStudy-bar">
+            <img class="caseStudy-image" src="/imgs/derek/logos/aeromexico-blk.png"/>
```
WidgetsBurritos: Make sure we delete any unused images from the repo.  For example, you're only using `aeromexico-blk.png` but you've included `aeromexico.png` in the repo, despite not using it.

styleguide/_themes/derek/scss/snowflakes/homepage.scss
```diff
@@ -291,14 +291,41 @@
...
+  background: $white;
+  display: block;
+  margin-bottom: 1px;
+  padding: 20px 50px 150px 20px;
+  text-decoration: none;
+}
 
 .caseStudy-wrapper {
-  background-image: url('../imgs/derek/banners/caseStudy.jpg');
+  background-image: url('../imgs/derek/banners/customer-story-background.jpg');
```
WidgetsBurritos: So I know the tendency here is that we already have these classes defined so we should just modify them, however...  There is a chance that some of these CSS selectors are being used elsewhere on the site.  As a result we introduce risk.  

Instead, since we're talking about a completely brand new customer story region, we should be using brand new classes/selectors. 

For example see https://github.com/rackerlabs/zoolander/pull/592/files#diff-ef43fa620c83e18c2fa2e5ea7aea1db7R1

Instead of replacing the existing announcement code, I created all new selectors.  

We could potentially even use the same naming convention here. For example: `.rsHomepage-customerStories-wrapper`

We can then target some of this other css for removal at a later time.

styleguide/_themes/derek/scss/snowflakes/homepage.scss
```diff
@@ -331,17 +364,156 @@
 
 .caseStudy-text {
   font-family: $copy;
-  font-size: 1.75rem;
+  font-size: 1.25rem;
   letter-spacing: .5px;
-  line-height: 2.5rem;
+  line-height: 1.5rem;
+  word-spacing: .5px;
```
WidgetsBurritos: I know we're just following examples of code that was here already, but I'd like to avoid using pixels for word spacing.  For consistency, we should use `rem` for font sizes and then `em` for letter-spacing/word-spacing/etc....

styleguide/_themes/derek/scss/snowflakes/homepage.scss
```diff
@@ -330,7 +330,7 @@
 }
 
 .caseStudy-text {
-  font-family: $copy;
+  font-family: "Open Sans","Helvetica Neue",Helvetica,Arial,sans-serif;
```
WidgetsBurritos: this got reverted to the wrong thing.  Can you make sure this is set to `$copy`?

styleguide/derek/incubation/homepage.ejs
```diff
@@ -5,6 +5,9 @@
   'body': 'Redhat names Rackspace North America Managed Service Partner of the Year'
 }) %>
 
+<!-- Case Study -->
```
WidgetsBurritos: Will you move this comment inside the partial file, and then maybe rename it "Customer Story River"?

styleguide/_themes/derek/scss/snowflakes/homepage.scss
```diff
@@ -342,7 +341,6 @@
   width: 250px;
 }
 
-
```
WidgetsBurritos: Would you mind just dropping this file from the PR altogether?  Nothing ended up changing here, so I'd prefer just leaving the file untouched.

styleguide/derek/pattern-components/cust-story-river/_cust-story-river.scss
```diff
@@ -0,0 +1,163 @@
...
+
+  .rsHomepage-customerStories-wrapper,
+  .rsHomepage-customerStories-main-image {
+    background-image: url('../imgs/derek/banners/customer-story-background-992.jpg');
+  }
+}
+// 1279 -  992px
+@media screen and (max-width: $screen-md-max) and (min-width: $screen-md) {
+  .rsHomepage-customerStories-wrapper,
+  .rsHomepage-customerStories-main-image {
```
WidgetsBurritos: Since `.rsHomepage-customerStories-main-image` isn't visible in this viewport let's drop `.rsHomepage-customerStories-main-image` here.

styleguide/derek/pattern-components/cust-story-river/_cust-story-river.scss
```diff
@@ -0,0 +1,163 @@
...
+    background-image: url('../imgs/derek/banners/customer-story-background-1280.jpg');
+  }
+}
+// 1280px+
+@media screen and (min-width: $screen-lg) {
+  .rsHomepage-customerStories-wrapper,
+  .rsHomepage-customerStories-main-image {
```
WidgetsBurritos: Since `.rsHomepage-customerStories-main-image` isn't visible in this viewport let's drop `.rsHomepage-customerStories-main-image` here.

styleguide/derek/pattern-components/cust-story-river/_cust-story-river.scss
```diff
@@ -0,0 +1,163 @@
...
+    padding-bottom: 60px;
+  }
+
+  .cs-story-box0,
+  .cs-story-box2 {
+    border-right: 1px solid $gray-light;
+  }
+
+  .rsHomepage-customerStories-wrapper,
+  .rsHomepage-customerStories-main-image {
```
WidgetsBurritos: Since we're relying on `.rsHomepage-customerStories-main-image` in this viewport, we should be safe to drop `.rsHomepage-customerStories-wrapper` here.

styleguide/derek/pattern-components/cust-story-river/_cust-story-river.scss
```diff
@@ -0,0 +1,163 @@
...
+@media screen and (max-width: $screen-sm-max) and (min-width: $screen-sm) {
+  .rsHomepage-customerStories-main-image {
+    display: block;
+  }
+
+  .rsHomepage-customerStories-wrapper {
+    background-image: none;
+    background-repeat: unset;
+    background-size: 0;
+  }
```
WidgetsBurritos: Since we're relying on `.rsHomepage-customerStories-main-image` in this viewport, we should be safe to drop `.rsHomepage-customerStories-wrapper` here.

styleguide/derek/pattern-components/cust-story-river/_cust-story-river.scss
```diff
@@ -0,0 +1,163 @@
...
+    padding-top: 10px;
+  }
+
+  .rsHomepage-customerStories-box {
+    margin-bottom: 0;
+    padding: 30px;
+  }
+
+  .rsHomepage-customerStories-wrapper,
+  .rsHomepage-customerStories-main-image {
```
WidgetsBurritos: Since we're relying on `.rsHomepage-customerStories-main-image` in this viewport, we should be safe to drop `.rsHomepage-customerStories-wrapper` here.

styleguide/derek/pattern-components/cust-story-river/cust-story-river.ejs
```diff
@@ -0,0 +1,29 @@
+<div class="rsHomepage-customerStories-wrapper">
+    <div>
```
WidgetsBurritos: Since we're not assigning a class or anything here I don't believe we need this `<div>` here.

styleguide/derek/incubation/homepage.ejs
```diff
@@ -5,6 +5,8 @@
   'body': 'Redhat names Rackspace North America Managed Service Partner of the Year'
 }) %>
 
+<%- partial("../pattern-components/cust-story-river/cust-story-river.ejs") %>
```
WidgetsBurritos: While we're leaving the CSS untouched we should still remove the legacy case study HTML from this page as this should show us what the new home page is to look like.

styleguide/derek/pattern-components/cust-story-river/_cust-story-river.scss
WidgetsBurritos: ^ I saw you removed all the other instances, we shouldn't need this one here either.

PR #597, opened 2018-11-27
chore(svg): get rid of unnecessary tags

WidgetsBurritos:
As it turns out we don't need all those tags in the rackspace SVG.  This should eliminate some of the noise.


runrog:
👍 LGTM

PR #599, opened 2018-11-29
MTE-3363: Home Page Banner Dev

xinbin:
https://jira.rax.io/browse/MTE-3363


xinbin:
Something is wrong with the homepage-banner-768.jpg image. For some reason it won't load. Not sure what the problem is:
issue_768_img

PR #600, opened 2018-11-29
MTE-3367: Add new brand challenges river

brantdarilek:


styleguide/derek/pattern-components/challenges-river/challenges-river.ejs
```diff
@@ -0,0 +1,69 @@
...
+      title: 'Consolidate data centers',
+      copy: 'Replace your aging hardware infrastructure with powerful modern architecture.'
+    },
+    {
+      shape: '<?xml version="1.0" encoding="utf-8"?><svg class="rackshape" version="1.1" id="Layer_1" x="0px" y="0px" width="201px" height="201px" viewBox="0 0 132 132" style="enable-background:new 0 0 201 201;" xmlns="http://www.w3.org/2000/svg"><style type="text/css">.st0{fill:none;stroke:#EB0000;stroke-miterlimit:10;}</style><g transform="matrix(1, 0, 0, 1, -34.5, -34.5)"><circle class="st0" cx="100.5" cy="68.5" r="32"/><circle class="st0" cx="100.5" cy="132.5" r="32"/><circle class="st0" cx="132.5" cy="100.5" r="32"/><circle class="st0" cx="68.5" cy="100.5" r="32"/></g></svg>',
```
WidgetsBurritos: Can we pull out the `<?xml version="1.0" encoding="utf-8"?>` tags here?   Turns out we don't need them in with embedded svgs.

styleguide/derek/pattern-components/challenges-river/_challenges-river.scss
```diff
@@ -0,0 +1,87 @@
...
+  }
+}
+
+.challenges-item {
+  @include make-xs-column(10);
+  @include make-md-column(8);
+
+  margin-bottom: 50px;
+
+  .rackshape {
```
WidgetsBurritos: Instead of nesting here can we not just define a top-level class called `.challenges-itemRackShape` or something like that?

styleguide/derek/pattern-components/challenges-river/_challenges-river.scss
```diff
@@ -0,0 +1,87 @@
...
+  .challenges-description {
+    padding-right: 0;
+  }
+}
+
+@media only screen and (max-width: 500px) {
+
+  .challenges-wrapper {
+    .challenges-descriptionRow {
+      .challenges-item {
```
WidgetsBurritos: is three-layers deep of nesting necessary here?

styleguide/derek/pattern-components/challenges-river/_challenges-river.scss
```diff
@@ -0,0 +1,87 @@
...
+  margin-top: 0;
+}
+
+@media screen and (max-width: $screen-xs-max) {
+  .challenges-description {
+    padding-right: 0;
+  }
+}
+
+@media only screen and (max-width: 500px) {
```
WidgetsBurritos: why are we using a non-standard breakpoint here?

brantdarilek: @WidgetsBurritos This breakpoint was done previously for the platforms river( https://github.com/rackerlabs/zoolander/blob/master/styleguide/derek/pattern-components/platforms/_platforms.scss#L112 ). Since the challenges section is very similar to/a replacement for platforms, I replicated it for this pull request. At 767px it looked odd to me to have the wider "challenges".

I personally think it looks better with the custom breakpoint, but I'm not married to it. If design/dev feels its better to not have it, thats fine. 

I've gone ahead and made a new branch that can be used for comparison.
http://mte-3367-challenge-river-wide-xs.styleguide.dev.website.rackspace.com/derek/incubation/homepage

brantdarilek: Sure thing. Not sure if you want to discuss with @runrog and Mike. But let me know. Like I said, not married either way.

PR #604, opened 2018-12-04
Met 3186 change resource logo

xinbin:
Update logo for resource center


styleguide/derek/pattern-components/resourceCenter-resource/refresh/_resourceCenter-resource.scss
```diff
@@ -69,7 +69,7 @@
 }
 
 .resource-logo {
-  background-image: url('../imgs/derek/rs-logo.svg');
+  background-image: url('../imgs/derek/rs_wordmark_white.svg');
   background-position: left 50%;
   background-repeat: no-repeat;
   background-size: 100% 79%;
```
WidgetsBurritos: I think we need to adjust the size and position a little bit.  

This is what is currently on production: 



This is what you have: 


Without fanatiguy it should probably align on the left.  Also the presence of fanatiguy made the text smaller previously.  Without it the text looks a little big. 


Let's try this: 
```
background-position: 39% 50%;
background-repeat: no-repeat;
background-size: 80% 63%;
```
This ends up looking like this:

PR #613, opened 2018-12-11
MTE-3477: Fix the landing page callout z-index

WidgetsBurritos:
This PR fixes the landing page call out z-index value.


runrog:
👍  That's a HUGE change but LGTM

PR #622, opened 2018-12-17
feat(style): making challenges items links

runrog:


styleguide/derek/pattern-components/challenges-river/_challenges-river.scss
```diff
@@ -22,11 +23,42 @@
...
+
 .challenges-item {
   @include make-xs-column(10);
   @include make-md-column(8);
+  height: 100%;
+  padding-bottom: 10px;
+  padding-top: 10px;
+  transition-duration: .2s;
+
+  svg {
```
WidgetsBurritos: instead of using the `svg` selectors here can we use `.challenges-itemRackShape` selector instead?

styleguide/derek/pattern-components/challenges-river/_challenges-river.scss
```diff
@@ -22,11 +23,42 @@
...
+    transition-duration: .2s;
+  }
 
-  margin-bottom: 50px;
+  &:hover {
+    box-shadow: 0 10px 50px -9px $black-shadow-darkest;
+
+    svg {
```
WidgetsBurritos: Same here

styleguide/derek/pattern-components/challenges-river/_challenges-river.scss
```diff
@@ -22,11 +23,42 @@
...
   @include make-xs-column(10);
   @include make-md-column(8);
+  height: 100%;
+  padding-bottom: 10px;
+  padding-top: 10px;
+  transition-duration: .2s;
+
+  .challenges-itemRackShape {
+    transition-duration: .2s;
+  }
```
WidgetsBurritos: Ehhh sorry, one last thing here. We should be able to move this `transition-duration` to the non-nested version of this selector (line 71).  I understand the need to keep it nested for the actual `transform` below due to it responding to the parent's hover state, but as far as the duration itself, it should be fine to live at the top-level.

PR #630, opened 2019-01-10
feat(style): adding new styling for interior event pages

runrog:
Adding new events pages look to current pages.

PR #626, opened 2018-12-19
2.0.x

WidgetsBurritos:
This won't actually get merged. Just doing this to generate a PR builder branch.

PR #631, opened 2019-01-10
Update node-sass version. maybe.

xinbin:
terst

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "karma-phantomjs-launcher": "^1.0.4",
     "karma-sinon-chai": "^1.2.4",
     "mocha": "^3.2.0",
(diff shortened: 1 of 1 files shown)


WidgetsBurritos:
Unfortunately it's not that simple.  There were some issues between this and the version of node that Drupal is pulling in that's causing this issue. This was fixed in the brand work, but since that got delayed we need to fix it on the master branch too.  See package.json changes here (excluding fira-sans): 
https://github.com/rackerlabs/zoolander/pull/626/commits/ed38f7c804ac2c9cc5f1ea7e850d010724f9b904#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

WidgetsBurritos:
Closing in favor of what @runrog did in #630

PR #633, opened 2019-01-14
refactor(stories): Change out filter for existing resources filter.

xinbin:
Change out the filter used on the customer story page for the one used on the resource center..


xinbin:
[run ci]

PR #645, opened 2019-03-14
MTE-3912: Spanish Navigation issue with tablet size.

pobtastic:
Adjusts the medium font-size so that the navbar and search icon align on the medium breakpoint.

https://jira.rax.io/browse/MTE-3912

### DIFF
--- a/styleguide/derek/global-components/mainnav/_mainnav.scss
+++ b/styleguide/derek/global-components/mainnav/_mainnav.scss
@@ -180,6 +180,13 @@
   }
 }
 
+@media only screen and (max-width: $screen-md-max) {
+  .navbar-nav {
+    .navbar-topLink {
+        font-size: $navigation-font-size-medium;
+    }
+  }
+}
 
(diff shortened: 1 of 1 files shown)


WidgetsBurritos:
@pobtastic ,

So we either need to also define a mid range breakpoint: 
```css
@media only screen and (min-width: $screen-md-min) and (max-width: $screen-md-max) {
  .navbar-nav {
    .navbar-topLink {
        font-size: $navigation-font-size-medium;
    }
  }  
}
```
Or we should update your change to use `screen-md-max` instead of `screen-sm-max`. 
```css
@media only screen and (max-width: $screen-md-max) {
  .navbar-nav {
    .navbar-topLink {
        font-size: $navigation-font-size-medium;
    }
  }  
}
```

The first option is useful if we want different font sizes for small -vs- medium. 
The second change is preferred if we just use the same font size for each.

runrog:
👍 LGTM @WidgetsBurritos ?

WidgetsBurritos:
+1 Ship it

PR #648, opened 2019-03-21
fix(nav): adding toggle for off hover on dropdown triggers

runrog:
This PR fixes the nav issues where the drop down triggers were not hiding submenu's when hovering off of them.
nav-hover-fix


pobtastic:
LGTM 👍

PR #658, opened 2019-04-03
feat(templates): Thank you lp templates.

xinbin:
There are 3 ty page variants. Created 3 pages.


styleguide/derek/pattern-components/webinar-item/_webinar-item.scss
```diff
@@ -44,12 +44,22 @@
...
-.rsWebinarTopic-item-link a {
-  color: $rackspace-red;
-  font-weight: $font-weight-regular;
-  text-decoration: underline;
+a {
+  .rsWebinarTopic-item-link {
+    color: $rackspace-red;
+    font-weight: $font-weight-regular;
+    text-decoration: underline;
+  }
```
WidgetsBurritos: What is this change for? Are we using this anywhere in this PR? It doesn't seem relevant to this thank you pages, and also seems like it's incorrect based on the markup on webinars anyway: 
```
<span class="rsWebinarTopic-item-link"><a href="#" class=""><%= linktext %></a></span>
```

styleguide/derek/pattern-components/webinar-item/_webinar-item.scss
```diff
@@ -44,12 +44,22 @@
...
+  .rsWebinarText-content {
+    font-weight: $font-weight-regular;
+  }
+}
+
+.rsTy-button {
+  background-color: $rackspace-red;
+  color: $white;
+  text-decoration: none;
 }
```
WidgetsBurritos: This shouldn't be in the webinar css file.  It's not related to that.   Also, I think CTAs are a case where we can stick to non-maintainable classes since we tend to use those consistently across a variety of applications.   I think we can just stick with `<a href="" class="button lead">Text<a>` here.

styleguide/derek/templates/landing-page/_partials/ty-text-section.ejs
```diff
@@ -0,0 +1,14 @@
+<!-- :(  -->
+<br>
+<div class="row">
+  <div class="col-sm-7">
+    <h1>Thank You!</h1>
+    <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.</p>
+    <a href="#" class="button rsTy-button">Download NOW!!!</a>
```
WidgetsBurritos: See comment about CTA above.

styleguide/derek/templates/index.jade
```diff
@@ -87,3 +85,10 @@
       a.red(href="derek/templates/landing-page/landingpage-tradeshow-event") Tradeshow Event Landing Page
     li
       a.red(href="derek/templates/landing-page/landingpage-webinar") Webinar Landing Page
+    li
+      a.red(href="derek/templates/landing-page/landingpage-thankyou1") Thank You v1
+    li
+      a.red(href="derek/templates/landing-page/landingpage-thankyou2") Thank You v2
+    li
+      a.red(href="derek/templates/landing-page/landingpage-thankyou3") Thank You v3
+
```
WidgetsBurritos: You must have missed my comment yesterday about not needing to actually produce all three variants here. Let's just leave the 2 column version and drop the other two, otherwise it's too noisy and if we have to make a change to one, we'll have to make a change to all three.  

We already have coverage in other ways to see the related resource variants anyway, so we don't need it here.

styleguide/derek/pattern-components/webinar-item/_webinar-item.scss
xinbin: There are lint errors that I was trying to quiet. and yeah I now see the change while it quiets the lint error is not correct.

styleguide/derek/pattern-components/webinar-item/_webinar-item.scss
xinbin: removing. I wasn't aware of the existing class and I didn't know where to otherwise add the css to.

xinbin:
As per the empty files. I was trying to rectify that now and delete them. I thought I did delete them. I don't know why they're showing the way they are in github. I'll try `git rm`

PR #662, opened 2019-04-17
feat(new pattern): multiple agendas per day pattern for events

runrog:


xinbin:
merging for release https://jira.rax.io/browse/MTE-4169

PR #664, opened 2019-04-23
fix(agenda style): fixing mobile issues for some agenda patterns

runrog:


styleguide/derek/view-templates/events/_events.scss
```diff
@@ -543,12 +543,31 @@
   }
 }
 
-// Mobile Behavior.
-@media screen and (min-width: $screen-xs-min) and (max-width: $screen-sm-min - 1) {
+@media screen and (max-width: $screen-sm-min - 1) {
```
WidgetsBurritos: We probably should use `$screen-xs-max` instead:
> $screen-xs-max:              ($screen-sm-min - 1) !default;

styleguide/derek/view-templates/events/_events.scss
```diff
@@ -543,12 +543,31 @@
...
+
+  .rsEvent-multiAgenda-info {
+    padding-left: 0;
+    padding-top: 15px;
+    width: 100%;
+  }
+}
+
+// Mobile Behavior.
+@media screen and (min-width: $screen-xs-min) and (max-width: $screen-sm-min - 1) {
```
WidgetsBurritos: same here

PR #680, opened 2019-05-08
MTE-4372: Add trigger-optimize class

WidgetsBurritos:
This PR adds the trigger-optimize class for dynamically triggering google optimize.


WidgetsBurritos:
Not doing this after all. Found a better solution

PR #684, opened 2019-06-14
fix(travis): Resolve test failure

WidgetsBurritos:
This PR resolves the travis test failures.

Changes made:

1. Made a slight modification to `tracking.spec.js` which was the actual culprit.  The form submission was causing a page reload, so I force the submit action to return false now instead so it will only listen to javascript events. 

2. Installed and configured the `karma-verbose-reporter` package to provide a little bit more visual feedback into the test pass/failures
image

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -62,7 +62,7 @@ module.exports = function (config) {
     // test results reporter to use
     // possible values: 'dots', 'progress'
     // available reporters: https://npmjs.org/browse/keyword/karma-reporter
-    reporters: ['progress'],
+    reporters: ['verbose'],
 
 
     // web server port
--- a/styleguide/derek/global-components/tracking/tracking.spec.js
+++ b/styleguide/derek/global-components/tracking/tracking.spec.js
@@ -445,7 +445,7 @@ describe('Zoolander Tracking Module', () => {
     });
 
     it('should track resource filtering products', () => {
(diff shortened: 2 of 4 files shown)


karma.conf.js
```diff
@@ -62,7 +62,7 @@ module.exports = function (config) {
     // test results reporter to use
     // possible values: 'dots', 'progress'
     // available reporters: https://npmjs.org/browse/keyword/karma-reporter
-    reporters: ['progress'],
+    reporters: ['verbose'],
```
xinbin: 👌

styleguide/derek/global-components/tracking/tracking.spec.js
```diff
@@ -445,7 +445,7 @@ describe('Zoolander Tracking Module', () => {
...
+      $('body').append('<form class="rsFilter-form" onsubmit="return false;"><select name="product"><option value="96" selected="selected">Public Cloud</option><option value="101">-Fanatical Support for AWS</option><input type="checkbox" name="content_type[]" id="edit-content-type-case-study" value="case_study" checked="checked" class="rsFilter-checkbox"><input type="checkbox" name="content_type[]" id="edit-content-type-code-repository" value="code_repository" class="rsFilter-checkbox"><input type="checkbox" name="content_type[]" id="edit-content-type-ebook" value="ebook" checked="checked" class="rsFilter-checkbox"></form>');
```
xinbin: 🧠

PR #694, opened 2019-06-27
fix(region selector): moving selector when in tiny viewport

runrog:


styleguide/derek/global-components/modal/_region-select.scss
```diff
@@ -128,3 +128,27 @@
     column-gap: 15px;
   }
 }
+
+@media only screen and (max-width: $screen-xs-min - 1) {
```
WidgetsBurritos: I don't think we want to stop at 479px here.   Some android devices use 480px and there's still some overlap there: 
image

I'd say either set it to `$screen-xs-min` (480px) or `$screen-xs-max` (767px).

styleguide/derek/global-components/modal/_region-select.scss
```diff
@@ -128,3 +128,9 @@
     column-gap: 15px;
   }
 }
+
+@media only screen and (max-width: $screen-xs-max) {
+  .basement-container {
+    padding-bottom: 35px;
```
WidgetsBurritos: So looking at production right now the chat button takes up a good 65px or so.  So maybe we just bump the padding to 70px instead?

xinbin:
looks good to me
Screen Shot 2019-06-28 at 12 35 26 PM

styleguide/derek/global-components/modal/_region-select.scss
pobtastic: He made it 75px ~ and I think this looks fine.

PR #711, opened 2019-08-29
MTE-4907: Resolve duplicate banner load issue

WidgetsBurritos:
This PR should resolve the duplicate banner issue on the homepage for browsers wider than 992px

Left side is production, right side is my branch: 
image


runrog:
👍 LGTM verified

PR #712, opened 2019-09-04
fix(tables): fix for plugin adding multiple headers for mobile tables

runrog:
Table plugin now checks if the mobile title is already there before it tries to add it.


pobtastic:
LGTM 👍

PR #723, opened 2019-11-20
chore(solve): Switch banners to <picture> tags

WidgetsBurritos:
This PR changes banners from inline css to `<picture>` tags to allow for use of responsive images inside drupal.


runrog:
Can you give .rsTl-feature-header a max height. something like 500px should do. The image get's way too massive on wider screens.

runrog:
+1

PR #725, opened 2019-11-26
feat(quote): Change quote to live in blockquote element.

xinbin:
https://jira.rax.io/browse/MTE-5049

I tried to make the quote look like the attachment in the ticket.
Also I completely removed the attribution info, or rather the way I understood it the part that indicates whom the quote is from...
I changed the quote characters.
I shortened the red dividers to match the image in the ticket

### DIFF
--- a/styleguide/derek/templates/thought-leadership-overview/article.ejs
+++ b/styleguide/derek/templates/thought-leadership-overview/article.ejs
@@ -47,19 +47,12 @@
     </div>
   </div>
 
-  <hr class="rsTl-divider" />
-
   <div class="rsTl-article-row">
-    <div class="rsTl-column-quote">
+    <blockquote>
       Nam libero tempore, cum soluta nobis est eligendi optio cumque nihil impedit quo minus id quod maxime
       placeat facere possimus, omnis voluptas assumenda est, omnis dolor repellendus.
-    </div>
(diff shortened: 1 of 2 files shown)


styleguide/derek/templates/thought-leadership-overview/article.ejs
```diff
@@ -47,19 +47,15 @@
     </div>
   </div>
 
-  <hr class="rsTl-divider" />
-
+  <hr class="rsTl-divider-top-quote" />
   <div class="rsTl-article-row">
-    <div class="rsTl-column-quote">
+
+    <blockquote class="rsTl-column-quote">
```
WidgetsBurritos: I know it's contrary to what we normally do but we actually want to style the `<blockquote>` element here without having to provide a class. The reason is we're going to have a lot of people editing content here, who won't know how to apply these classes. 

Also, we shouldn't require users to add `<hr>` tags either.  We can accomplish the same effect by applying a `border-top` and a `border-bottom` to the `<blockquote>` tag. 

The caveat, is we only want this apply to blockquotes for thought leadership nodes.  To do that, we'd probably want to target something like this: 

```
.rsTl-article-row blockquote
```

runrog:
second commit LGTM 👍

PR #728, opened 2019-12-03
feat(style): make thought leadership articles links

runrog:


WidgetsBurritos:
My only question, is should there be extra padding around the text here, so that it's not butting right against the border? @mikerastiello thoughts?
image

WidgetsBurritos:
Also, it seems to me that the red "button" (which is kinda meaningless now), has its own hover state. Perhaps it should change colors when the whole container changes colors?

PR #729, opened 2019-12-03
MTE-5223: Resolve linkedin tracking issue

WidgetsBurritos:
JIRA: https://jira.rax.io/browse/MTE-5223

This PR resolves an issue with LinkedIn share buttons. See ticket for details, but essentially what was happening is at the point the `systemReady` event fired, the `IN-widget` class wasn't actually rendered yet. Looking through the [LinkedIn SDK docs](https://developer.linkedin.com/docs/js-sdk/api-reference/inauth_inevent_inui#content-par_longformtext_2), there does not appear to be an event that fires any later than `systemReady` so we can't rely on LinkedIn to know when that button is rendered. So what I've done instead is allowed for up to 300ms of time after `systemReady` for it to load, checking every 100ms. All of the tests I've done show it only needing one iteration through the interval, but I've allowed for 3 just in case. If it takes any longer than that, I think we're okay losing tracking here. 

This is definitely a hacky solution, but I don't see any way around it given the limitations of the LinkedIn API. 

Here's a screenshot of it working on a test environment: 
image

### DIFF
--- a/styleguide/derek/global-components/social/_social.scss
+++ b/styleguide/derek/global-components/social/_social.scss
@@ -19,7 +19,6 @@ $xing-green: #005a5f;
   padding: 0 0 0 2px;
   position: relative;
   text-decoration: none;
-  top: -1px;
   vertical-align: top;
   white-space: nowrap;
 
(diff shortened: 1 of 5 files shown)


pobtastic:
LGTM 👍

PR #731, opened 2019-12-09
MTE-5235: Fix overlay issue

WidgetsBurritos:
This PR fixes a bug where the solve nav overlaid the banner content.  The DOM structure in zoolander is slightly different in Drupal vs Zoolander, so this is a less specific approach.

Also the article pages themselves should not have the "banner" region be a link.  That's just for the overview app. So I changed from `<a>` to `<div>` in zoolander here to match what's in Drupal.


pobtastic:
LGTM

PR #732, opened 2019-12-09
MTE-5238: Remove underline on focus state

WidgetsBurritos:
This PR fixes a bug with the focus state on solve links.


pobtastic:
LGTM

PR #733, opened 2019-12-09
MTE-5235: Fixes overlap issue on solve app

WidgetsBurritos:
#731 didn't completely solve the problem. When used in combination with the solve app itself (which includes redraws), the element containing the class `.rsTl-feature-header` is redrawn, which means predefining a variable with it, doesn't work. Instead, we have to do a lookup each time. Also I bumped the timeout to a smaller window, to minimize time to redraw. Finally, I had to force a resize on init due to the way the SPA loads.


pobtastic:
Ah yeah, I understand ~ LGTM 👍

PR #735, opened 2019-12-10
MTE-5244: Make only certain things linkable

WidgetsBurritos:
JIRA: https://jira.rax.io/browse/MTE-5244
Relates to: rackerlabs/solve#19

This PR makes only certain things clickable.


runrog:
👍 LGTM

PR #739, opened 2019-12-12
MTE-5245: Set max-width on content container

WidgetsBurritos:
JIRA: https://jira.rax.io/browse/MTE-5245

This PR sets a max container of 970px on the content portions of the article.


pobtastic:
LGTM

PR #740, opened 2019-12-16
MTE-5260: Make sponsor logo optional

WidgetsBurritos:
JIRA: https://jira.rax.io/browse/MTE-5260

This PR makes the sponsor logo optional. There is a www PR to follow as well.


pobtastic:
Shouldn’t we do this in the template so Javascript being enabled isn’t a requirement for the styling?

WidgetsBurritos:
@pobtastic the `.rsTl-nav-logoWrapper` element doesn't get created unless javascript is enabled. And for what it is, I think it's acceptable for it to be missing for the few users with javascript disabled.

pobtastic:
Perfect! Thanks 👌 LGTM +1

PR #748, opened 2019-12-30
fix(style): adding better contrast to footer text

runrog:


xinbin (approved):
err looks legit to me? +1

PR #745, opened 2019-12-20
MTE-5273: Update packages for node 10-13 support

WidgetsBurritos:
JIRA: https://jira.rax.io/browse/MTE-5273
www: rackerlabs/www#469
Solve: rackerlabs/solve#29

This PR updates all the packages to provide node 10/12 support.

PR #749, opened 2019-12-31
chore(node): Update packages

WidgetsBurritos:
This PR updates a few packages I missed in #745 
- `clean-css` and `cli` needed to be updated explicitly due to minor security issues with the versions that come packaged with harp. 
- `sinon` just needs to be updated to the latest version.


runrog:
👍 LGTM

PR #755, opened 2020-01-23
fix(style): icon font size for key offers

runrog:
This looked to be more of an issue with the key offers widget.

### DIFF
--- a/styleguide/_themes/global/scss/fonts.scss
+++ b/styleguide/_themes/global/scss/fonts.scss
@@ -157,7 +157,9 @@ i {
   line-height: 1;
   speak: none;
   text-transform: none;
+}
 
+span {
   &.rsweb {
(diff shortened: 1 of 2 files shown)


WidgetsBurritos (changes requested):
It's believe zoolander only, but can fix this reference too? 
https://github.com/rackerlabs/zoolander/blob/3.0.x/styleguide/_themes/global/scss/fonts.scss#L151-L169
```scss
i {
  ...

  &.rsweb {
    font-family: 'rswebfonts';
    font-size: 4.5em;
  }

  ...
}
```

runrog:
@WidgetsBurritos ah that's what it was. My bad, didn't see that.

WidgetsBurritos (approved):
👍 LGTM

PR #759, opened 2020-02-03
feat(style): new exposed form styles option

runrog:


styleguide/_themes/global/scss/forms.scss
```diff
@@ -228,3 +228,232 @@ table {
     }
   }
 }
+
+// exposed form overrides
+$exposed-form-bg: #003482;
+
+// sass-lint:disable-block no-ids
+#block-webform {
```
WidgetsBurritos: Can we change this to: `.rs-webform-embedded`?

I was able to tweak the HTML to this: 
```
<div class="rs-webform-block rs-webform-embedded" id="block-webform-block">
```

That `.rs-webform-embedded` class is specific to this particular block and you can see it's at the same level as that `id`.

styleguide/derek/global-theme/forms.ejs
```diff
@@ -77,6 +77,80 @@
   </form>
 </div>
 
+<div class="row half-padding">
+  <h5>Exposed Form</h5>
+  <div id="block-webform">
```
WidgetsBurritos: Change this to `<div class="rs-webform-block rs-webform-embedded" id="block-webform-block">` to match what we have in production and then target off of `rs-webform-embedded`

runrog:
@WidgetsBurritos that error message injecting to the top of the form threw a major wrench. Now we have to target the other generic `.form-items` uglier. `.form-item:nth-type-of(num)` doesn't work as expected so I had to go with a nested sibling selector. The other option would be js. Rather not go that route but it should be working now.

WidgetsBurritos (approved):
👍 LGTM

PR #765, opened 2020-02-13
MTE-5396 feat(fonts): adding in 500 weight fonts back

runrog:
This basically reverts this PR: https://github.com/rackerlabs/zoolander/commit/ed1235ae5323cd929b1f92a81b262081cbbe7e55

However adding in the 500 weight.


xinbin (approved):
+1 I have in fact verified that 600 is now 500.

WidgetsBurritos (changes requested):
See feedback in slack

PR #773, opened 2020-03-12
MTE-5371: Remove problematic <p> tags

WidgetsBurritos:
This PR removes problematic `<p>` tags and adds a top/bottom margin to the CTA so things are spaced out correctly.


xinbin (approved):
+1

PR #776, opened 2020-03-20
MTE-5486: Update minimist package

xinbin:
https://jira.rax.io/browse/MTE-5486


WidgetsBurritos (approved):
👍 Let's see what happens

PR #777, opened 2020-03-25
MTE 5473 - fix(styles): adding fallback twitter share styling

runrog:


WidgetsBurritos:
Closing since this is covered in rackerlabs/www#600

PR #798, opened 2020-05-26
feat(banner): new homepage banner

runrog:
Pending review


styleguide/derek/pattern-components/page-banner/_page-banner.scss
```diff
@@ -1,6 +1,6 @@
 
 .pageBanner-wrapper {
-  background-image: url('../imgs/derek/banners/homepage-banner-480.jpg');
+  background-image: url('../imgs/derek/banners/hp-480.jpg');
```
WidgetsBurritos: @runrog instead of overriding existing classes, can we create new classes for this? I want to be able to deploy the zoolander changes out without any existing homepage elements.

WidgetsBurritos (changes requested):

styleguide/derek/pattern-components/page-banner/_page-banner.scss
runrog: @WidgetsBurritos wouldn't this create more work? new classes mean we now have to update our templates.

WidgetsBurritos (approved):
👍 Works for me

WidgetsBurritos (approved):
👍 LGTM

PR #800, opened 2020-06-02
MTE-5721: Logo ticker

WidgetsBurritos:
JIRA: https://jira.rax.io/browse/MTE-5721

This PR adds a logo ticker


styleguide/derek/pattern-components/logo-ticker/logo-ticker.ejs
```diff
@@ -0,0 +1,77 @@
...
+
+<div class="logoTicker-wrapper">
+  <div class="logoTicker">
+    <% for(var i=0; i<logos.length; i++) {%>
+      <%- partial("_partials/logo", logos[i]) %>
+    <% } %>
+
+    <% for(var i=0; i<logos.length; i++) {%>
+      <%- partial("_partials/logo", logos[i]) %>
+    <% } %>
```
WidgetsBurritos, line 75: Note: Because of the way the calculations work, we have to load in all logos twice. I have confirmed multiple `` references to the same resource only incur a single connection (at least in firefox). Also, when we put this in the general in Drupal we can add the necessary lazy loading classes and change `src` to `data-src`. I only didn't do it here because we don't have all the lazy-loading facilities in zoolander.

PR #801, opened 2020-06-02
MTE-5688: Add "Alert" Bar Functionality

xinbin:
https://jira.rax.io/browse/MTE-5688


WidgetsBurritos:
Closing since this has evolved in the CMS a bit since

PR #807, opened 2020-06-12
MTE-5790: Add content activation tracking

WidgetsBurritos:
Opening against the main branch as well, just so this doesn't get lost in the shuffle.


xinbin (approved):
`+ 1`

PR #814, opened 2020-07-10
fix(ca): Adjust trigger style to allow for inline content

WidgetsBurritos:
This PR removes the `.rs-gatedContent-trigger` style so content can be used inline. 

I also modified the sample markup to show triggers in a different order, and across multiple HTML elements, just to illustrate that it is possible to do so.


pobtastic:
Perfect! +1

PR #816, opened 2020-07-20
MTE-5843:  play button partially covered

xinbin:
https://jira.rax.io/browse/MTE-5843

Change summary:
Using the `resourcecenter-visualWrapper` instead to apply the fix. Added margin-bottom: 20px. It seems to be enough to resolve the issue.


WidgetsBurritos (approved):
👍  LGTM

PR #819, opened 2020-07-24
MTE-5907: Track clicking of nested elements

WidgetsBurritos:
JIRA: https://jira.rax.io/browse/MTE-5907

This PR fixes tracking classes, such that when nested elements are clicked, corrected tracking data is still attributed in GA.

### DIFF
--- a/styleguide/derek/global-components/tracking/tracking.js
+++ b/styleguide/derek/global-components/tracking/tracking.js
@@ -77,7 +77,7 @@ Zoolander.Tracking = (function Tracking($) {
         event: 'ga.event',
         eventCategory: 'Content Activation',
         eventAction: 'Modal Click',
-        eventLabel: $(e.target).data('target'),
+        eventLabel: $(e.currentTarget).data('target'),
         eventValue: '0',
         eventNonInteraction: 0,
       });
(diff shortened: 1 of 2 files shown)


styleguide/derek/global-components/tracking/tracking.js
```diff
@@ -197,14 +197,18 @@ Zoolander.Tracking = (function Tracking($) {
...
     $(document).on('click', '.track-signupMenu', (e) => {
-      const protocol = $(e.target).prop('protocol');
-      const hostname = $(e.target).prop('hostname');
-      const pathname = $(e.target).prop('pathname');
+      const protocol = $(e.currentTarget).prop('protocol');
+      const hostname = $(e.currentTarget).prop('hostname');
+      const pathname = $(e.currentTarget).prop('pathname');
+      const hash = $(e.currentTarget).prop('hash');
+      const port = $(e.currentTarget).prop('port');
+      const portStr = port ? `:${port}` : '';
```
WidgetsBurritos, line 205: Note: I did a little bit more than just switching to e.currentTarget here. Specifically, I added the ability to add the port and the URL hash to the tracked URL when applicable. The reason for that, is so that I could just attribute some links to http://localhost:9876 instead of actually clicking through to actual URLs. That was causing issue with karma.

xinbin (approved):
+1 looks good to me.

PR #821, opened 2020-07-27
fix(chat): adding css to hard seet bottom property for chat slider

runrog:


styleguide/derek/global-components/chat/chat.scss
```diff
@@ -1,3 +1,22 @@
...
+  _::-webkit-full-page-media, _:future, :root body:not([style*='padding-bottom: 0px']) .helpButton {
+    bottom: 168px !important;
+  }
+}
+
+@media screen and (min-width: 1600px) {
+  _::-webkit-full-page-media, _:future, :root body:not([style*='padding-bottom: 0px']) .helpButton {
+    bottom: 120px !important;
+  }
+}
```
WidgetsBurritos, line 22: Maybe we should just inject this style into the CMS block itself. I know @xinbin is working on fixing some stuff on here end.  Maybe she could dump it in there so we can verify?

runrog: @WidgetsBurritos Ok sounds good to me.

WidgetsBurritos:
This is not a thing

PR #823, opened 2020-08-18
MTE-5959: Fix Icons on LP Eyebrow

xinbin:
https://jira.rax.io/browse/MTE-5959

Change summary:
- this change attempts to hide the text in the element, but not the icon, of course when in table/mobile form only


WidgetsBurritos:
[run ci]

^ Testing that my changes work

PR #826, opened 2020-09-18
MTE-6114: Add Solve for Co-Branding

xinbin:
https://jira.rax.io/browse/MTE-6114


WidgetsBurritos:
This should be handled in www only like the other logos. See feedback in www PR.

PR #839, opened 2021-07-19
MTE-7205: Update landing page logo

WidgetsBurritos:
JIRA: https://jira.rax.io/browse/MTE-7205

Update landing page logo


WidgetsBurritos:
Mike signed off in JIRA

